PR #198, opened 2026-04-10
fix(ci): add --no-isolation to python -m build to fix semantic release

fgogolli:
## Description

The semantic release build was failing with `.venv/bin/pip: cannot execute: required file not found`.

**Root cause:** `python -m build` runs in isolation mode by default, which spawns a pip subprocess to install build dependencies into a temporary environment. In a uv-managed venv, `.venv/bin/pip` is a non-functional stub. When the `python-semantic-release` Action runs `make semantic-release-build` as a subprocess, `uv` is not on its PATH (Actions run in a sandboxed environment that doesn't inherit the runner's full PATH). So `run_tool.sh` falls through to the `.venv` branch, and `python -m build` then tries to call `.venv/bin/pip` which cannot execute.

**Fix:** Add `--no-isolation` to all `python -m build` invocations in `build.sh`. This uses the already-populated venv directly, bypassing the pip subprocess entirely. `setuptools` and `wheel` are already present from `uv sync --all-extras`.

This fix:
- Uses `$RUN_TOOL` consistently — no hardcoded `uv`
- Eliminates the pip subprocess entirely
- Works whether `uv` is on PATH or not
- Does not change the workflow, `deploy.mk`, or `run_tool.sh`

## Type of Change

## Checklist

## Reviewers
@awslabs/orb-maintainers

### DIFF
--- a/dev-tools/package/build.sh
+++ b/dev-tools/package/build.sh
@@ -29,27 +29,13 @@ if [ "$QUIET" = false ]; then
 fi
 rm -rf dist/ build/ -- *.egg-info/
 
-# Install build dependencies if needed
+# Verify build dependencies are present (declared in pyproject.toml dev deps)
 if [ "$QUIET" = false ]; then
     echo "INFO: Checking build dependencies..."
 fi
 if ! $RUN_TOOL python -c "import build" 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


kirillsc (approved):
lgtm

PR #211, opened 2026-04-30
fix: HF spec compliance, DTO consolidation, metadata/provider_data boundary enforcement

fgogolli:
## Summary

Comprehensive cleanup of the Machine/Request data pipeline, driven by a field-by-field audit against the IBM Symphony HostFactory 7.3.2 spec (verified verbatim from IBM docs).

- **HF spec violations fixed** — getReturnRequests was emitting the wrong response shape entirely; template attributes had raw ints instead of string-wrapped numerics; machine-level message could be empty when result=fail (spec mandates non-empty); spurious ncores/region keys removed
- **MachineDTO consolidated** — both query handlers were constructing DTOs inline, bypassing from_domain and silently dropping 6+ fields (cloud_host_id always null on the wire). Killed the long gate, made from_domain the single source of truth
- **RequestDTO cleaned** — wired first_status_check/last_status_check end-to-end, moved verbose to to_dict() parameter, deleted deprecated from_domain, removed HF-irrelevant fields from MachineReferenceDTO
- **metadata vs provider_data rule established** — architecture test enforces the boundary (0 violations remaining). AWS adapter migrates vcpus/availability_zone/region from metadata to provider_data with on-read legacy migration
- **Orchestration telemetry moved** — provider strategies no longer stamp application-level routing decisions; ProviderResult gains typed routing_info field, application layer stamps after call returns
- **Default scheduler surfaces rich data** — region, AZ, vcpus, health_checks, status check timestamps as ISO strings via REST and CLI
- **health_checks synthesized** from existing describe_instances data (state_reason/state_transition_reason) — zero extra API calls, no new IAM permissions

## Test plan

- [x] pyright: 0 errors, 0 warnings
- [x] pytest: 6790 passed, 85 skipped (pre-existing Docker/performance skips)
- [x] Architecture test: metadata violations baseline at 0
- [x] HF contract tests validate against hf_docs/input-output.md samples
- [x] Round-trip serialization tests for all new fields
- [ ] Manual: verify HF getReturnRequests response shape with a real Symphony deployment
- [ ] Manual: verify default scheduler CLI output shows new fields


kirillsc (approved):
lgtm

PR #215, opened 2026-05-13
fix(watch): preserve provider_data so AZ and vCPUs render in watch

kirillsc:
## Summary

The watch loop iterated `MachineReferenceDTO`, the HF-shaped projection
that deliberately drops `provider_data`. As a result watch rendered AZ
as `unknown` and vCPUs as `0` even when the AWS adapter had populated
them.

Two related root causes are fixed together because they share the same
symptom (lost `provider_data`) and the watch fix is only observable
once persistence stops dropping the field:

1. **Watch orchestrator** now side-loads `MachineDTO` via
   `ListMachinesQuery(lightweight=True)` and joins by `machine_id`.
   A new `lightweight` flag on `ListMachinesQuery` skips the
   per-machine AWS describe sync, mirroring
   `GetRequestQuery.lightweight`, so the 1 Hz watch loop does not
   amplify into N AWS calls per tick.

2. **Machine factories** in `RequestStatusManagementService` and
   `MachineSyncService` no longer drop `provider_data` from the input
   dict. Freshly provisioned machines now persist `availability_zone`,
   `vcpus`, and `region` as written by `AWSMachineAdapter`.

## Type of Change

## How Has This Been Tested?
- [x] Unit tests added/updated — `test_watch_request_status_orchestrator.py`
      covers the side-load + join path and the lightweight flag.
- [x] Integration test added — `test_machine_provider_data_roundtrip.py`
      round-trips `provider_data` through both factories and the JSON
      storage strategy.
- [x] Manual testing — verified `request watch` now shows real AZ and
      vCPUs end-to-end against AWS.

## Checklist
- [x] Code follows project style (ruff auto-format applied)
- [x] Self-review performed
- [x] No new warnings
- [x] Tests added that prove the fix
- [x] Existing tests pass locally
- [ ] Documentation update — N/A (internal behaviour, no public API change)
- [ ] CHANGELOG.md — N/A
- [ ] Version bump — N/A

## Performance Impact
- [x] No significant performance impact. The new `lightweight=True`
      path on `ListMachinesQuery` strictly avoids work (skips
      per-machine AWS describe), so the 1 Hz watch loop is cheaper,
      not more expensive.

## Security Considerations

## Reviewers
@awslabs/orb-maintainers


fgogolli (commented):
## Review: fix(watch): preserve provider_data so AZ and vCPUs render in watch

Solid fix. Addressing both drop sites — `RequestStatusManagementService._create_machine_aggregate` and `MachineSyncService._create_machine_from_processed_data` — in one PR eliminates the risk of either regressing independently. The `asyncio.gather` parallelisation in the watch orchestrator is the right call, and the comment about benign display-time skew is honest and helpful. The new integration test in `tests/integration/test_machine_provider_data_roundtrip.py` is the right level: it proves the factory methods and JSON storage layer together, which is exactly where unit tests would have missed this.

### Suggestions

- **(minor — nit)** `src/orb/application/services/orchestration/watch_request_status.py` — the inline comment ends with a stray semicolon: `…which is immutable post-launch;`. Drop the `;`.

- **(minor — risk)** `src/orb/application/services/orchestration/watch_request_status.py` — `ListMachinesQuery(request_id=..., lightweight=True, limit=1000)` hard-codes the pagination ceiling. If a fleet ever exceeds 1,000 machines the watch display will silently under-count vCPUs and AZ stats with no error signal. Either switch to `all_resources=True` (or an equivalent bypass-pagination flag) or add a `TODO` comment documenting why 1,000 is a safe upper bound for expected fleet sizes.

- **(low — future-proofing)** `src/orb/application/queries/machine_query_handlers.py:113` — the `lightweight` guard skips the AWS describe sync for any caller that passes `lightweight=True`, not only the watch path. If a future caller sets the flag for an unrelated reason it will silently skip live-sync. Renaming to `skip_provider_sync` would make the intent unambiguous. Not blocking for this fix, but worth a follow-up.

### One thing verified fine

`@pytest.mark` usage in `tests/integration/test_machine_provider_data_roundtrip.py` — `integration` is registered in `pyproject.toml` under `[tool.pytest.ini_options].markers`, so no `PytestUnknownMarkWarning` risk.

---

The storage fix and the watch orchestrator changes are independently correct. The `limit=1000` ceiling is the most actionable real risk before merge; everything else is non-blocking.

src/orb/application/services/orchestration/watch_request_status.py
```diff
@@ -47,8 +53,26 @@ async def _load_template_weights(self, template_id: str) -> tuple[dict[str, int]
...
-        result = await self._query_bus.execute(query)
+        # Fetch request (membership + lifecycle) and machine snapshots in parallel.
+        # awaits resolve at slightly different timestamps, which is benign for a
+        # display loop late arrivals just bucket as "unknown" for one tick.
+        # Refresh request: status drives the watch loop's terminal check; skip cache to see live transitions.
+        request_query = GetRequestQuery(
+            request_id=input.request_id, lightweight=False, skip_cache=True
+        )
+        # ListMachinesQuery(lightweight=True) skips the AWS describe sync; we only
+        # need provider_data already persisted on the Machine aggregate which is immutable post-launch;
```
fgogolli, line 64: Nit: drop the trailing `;` — Python comments don't use terminators.

fgogolli (commented):
Follow-up: per-line annotations for the items called out in the summary review.

src/orb/application/services/orchestration/watch_request_status.py
```diff
@@ -47,8 +53,26 @@ async def _load_template_weights(self, template_id: str) -> tuple[dict[str, int]
...
+        machines_query = ListMachinesQuery(
+            request_id=input.request_id, lightweight=True, limit=1000
```
fgogolli, line 66: `limit=1000` is a silent ceiling: any request with more than 1000 machines will produce under-counted vCPUs and AZ stats in the watch display with no error signal. Consider using `all_resources=True` (if `ListMachinesQuery` gains that field) or at minimum add a `# TODO: raises a ceiling` comment documenting the assumption that active requests stay well under 1000 machines.

src/orb/application/queries/machine_query_handlers.py
```diff
@@ -113,8 +113,13 @@ async def execute_query(self, query: ListMachinesQuery) -> list[MachineDTO]:
 
                 machine_dtos = []
                 for machine in machines:
-                    # Refresh running machines with live AWS state before building DTOs
-                    if machine.status.value == "running" and machine.request_id:
+                    # Refresh running machines with live AWS state before building DTOs.
+                    # Skipped when query.lightweight is True (polling callers).
+                    if (
+                        not query.lightweight
```
fgogolli, line 119: `skip_provider_sync` would be more self-documenting: the flag's only semantic is bypassing the live AWS describe sync, and a generic name like `lightweight` invites future callers to re-use it for unrelated purposes and accidentally skip the sync. Non-blocking — fine to defer.

PR #234, opened 2026-06-07
feat(oci): add config-driven OCI provider support

ncusato:
## Description
Adds config-driven OCI provider support for ORB, including OCI provider registration, OCI Compute strategy/handler support, template mapping/defaults, init-time OCI credential and infrastructure discovery, packaged OCI template examples, and OCI-specific documentation.

This keeps OCI setup file-driven through `--config` and `--provider`. It intentionally does not add `orb providers add/update --oci-*` CLI flags. AWS and HostFactory-facing diffs were restored to upstream parity.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependencies update

## Related Issues
Fixes #

## How Has This Been Tested?

Focused local validation:
- `git diff --check oci-fsi/main...HEAD`
- OCI/shared/AWS parity unit suite: `290 passed, 2 deselected`

Manual testing included OCI CLI credential verification and `orb init` OCI setup/discovery flow on Windows.

## Test Configuration
* Python version: 3.12.13
* OS: Windows
* AWS region: N/A
* Dependencies changed: No runtime dependency changes; package data now includes `orb.providers.oci.config/*.json`.

## Checklist
- [x] I have made corresponding changes to the documentation

## Additional Notes
Impact-zone cleanup was verified: the PR diff does not include AWS provider files, HostFactory scripts/transformations, `config/aws_templates.json`, or broad `src/orb/cli/args.py` parser changes.

Known local test environment note: `tests/unit/api/test_templates_router.py` was not run locally because this Windows venv does not have `fastapi` installed. Two HostFactory user-data tests were deselected due existing Windows-local baseline behavior around `~` expansion and unreadable file permissions.

N/A

## Performance Impact
- [ ] Performance improved

## Security Considerations
- [x] Security improved

Adds support for OCI instance principal/resource principal-style authentication paths so deployments can avoid local API key material where appropriate. OCI local/API key configuration remains config-driven.

## Dependencies
No new runtime dependencies. OCI JSON defaults/templates are included as package data.

## Migration Guide
N/A. This is additive and should not change existing AWS behavior.

## Deployment Notes
For OCI, configure ORB with `config/oci_config.local.example.json` for local profile auth or `config/oci_config.remote.example.json` for OCI Compute instance principal auth. Use existing `--config` and `--provider` flags.

## Reviewers


src/orb/providers/registration.py
```diff
@@ -12,9 +12,10 @@ def register_all_provider_types() -> None:
 
     register_aws_provider(registry)
 
-    # Future providers would be added here
-    # register_provider1_provider(registry)
-    # register_provider2_provider(registry)
+    # Register OCI provider
+    from orb.providers.oci.registration import register_oci_provider
+
+    register_oci_provider(registry)
```
fgogolli, line 18: Main now provides `_REGISTERED_PROVIDERS: list[str] = ["aws"]` and a dynamic importlib dispatch loop in `register_all_providers(container)` - this landed in #253 after the branch was cut. After rebasing, the only change needed here is `_REGISTERED_PROVIDERS = ["aws", "oci"]`. The dynamic loop then picks up `register_oci_provider` and `initialize_oci_provider` automatically by convention. Drop the imperative rewrite.

src/orb/providers/oci/handlers/oci_compute_handler.py
```diff
@@ -0,0 +1,545 @@
...
+
+        completed = subprocess.run(
+            cmd,
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        if completed.returncode != 0:
+            raise RuntimeError(
+                f"OCI CLI command failed ({completed.returncode}): {' '.join(cmd)}\n"
```
fgogolli, line 89: Two issues here. `' '.join(cmd)` includes the `--from-json` argument appended at line 79, which serialises the full launch payload - `metadata.ssh_authorized_keys` and `metadata.user_data` appear verbatim in the error string and in every log line at the call site. Redact: replace the value of `--from-json` with `'<redacted>'` when constructing the error. Separately, `subprocess.run` is blocking I/O called from `create_instances`, which is `async def`. Wrap with `await asyncio.to_thread(subprocess.run, ...)` - `aws_provider_strategy.py:471` shows the established pattern.

src/orb/providers/oci/strategy/oci_provider_strategy.py
```diff
@@ -0,0 +1,689 @@
...
+    def initialize(self) -> bool:
+        try:
+            self._logger.info("OCI provider strategy ready for region: %s", self._config.region)
+            self._initialized = True
+            return True
+        except Exception as exc:
+            self._logger.error("Failed to initialize OCI provider strategy: %s", exc, exc_info=True)
+            return False
+
+    async def execute_operation(self, operation: ProviderOperation) -> ProviderResult:
```
fgogolli, line 85: `OCIComputeHandler` has no `check_hosts_status` method. Main now expects this method to return `CheckHostsStatusResult(instances, fulfilment=ProviderFulfilment(state, ...))` from `src/orb/domain/base/provider_fulfilment.py` - this contract didn't exist when the branch was cut. `RequestStatusService` asserts this contract and raises at runtime on any OCI provisioning request. Please implement it after rebasing, following the `EC2FleetHandler` pattern in `src/orb/providers/aws/infrastructure/handlers/`.

src/orb/interface/init_command_handler.py
```diff
@@ -503,6 +523,51 @@ def _pick_region(regions: list[tuple[str, str]], default_region: str = "") -> st
...
+    provider_config["credential_source"] = selected_source
+    if selected_source == "profile":
+        profile = input("  OCI profile (DEFAULT): ").strip() or "DEFAULT"
+        provider_config["profile"] = profile
+    elif selected_source == "api_key":
+        requirements = _get_credential_requirements(provider_type)
+        for param, info in requirements.items():
+            if info.get("required_if_source") == "api_key":
+                prompt = f"  {info['description']}: "
+                provider_config[param] = input(prompt).strip()
```
fgogolli, line 544: `private_key_path` is stored directly from `input()` with no validation. Apply `os.path.expandvars`, `os.path.expanduser`, and `os.path.isfile` before writing to config - the same treatment used in `_collect_ssh_public_key` in `oci_provider_strategy.py:597-612`. Without this, a mistyped path is silently stored and only fails at the first subprocess call; on a shared host, an unvalidated path from user input reaches the OCI CLI as a credential argument.

tests/unit/providers/oci/test_strategy_compute_operations.py
```diff
@@ -0,0 +1,565 @@
...
+    assert result.data["effective_capacity_type"] == "ondemand"
+    assert result.data["fallback_attempted"] is True
+
+
+@pytest.mark.asyncio
+async def test_create_instances_cli_unavailable_outside_mock_returns_error() -> None:
+    strategy = _make_strategy()
+    handler = strategy._compute_handler
+    handler._oci_cli_available = False
+    handler._is_test_context = MagicMock(return_value=False)
```
fgogolli, line 367: Patching a `@staticmethod` via an instance attribute works when the call site uses `self._is_test_context()`, but silently stops working if any call site uses `OCIComputeHandler._is_test_context()` (class-qualified). This is a direct consequence of the test-context detection being in the production class. After refactoring to an injected `CommandRunner` protocol (see body comment), this test can use `monkeypatch.delenv("PYTEST_CURRENT_TEST", raising=False)` or monkeypatch the injected runner directly - both approaches are resilient to call-site changes.

fgogolli (commented):
OCI compute provider via `oci` CLI subprocess - a distinct integration approach from the boto3-based AWS provider. The overall structure is solid and the OCI-specific code in `providers/oci/` is well-scoped. The primary action is a rebase onto `dd4c3903a` (post-#253); several items below are conflicts that resolve on rebase, and a handful require active work in the OCI code itself.

**Rebase-alignment items**

Main now provides three domain files added after this branch was cut: `provider_fulfilment.py` (`ProviderFulfilment` + `CheckHostsStatusResult`), `operation_outcome.py` (`Accepted | Completed | RequiresFollowUp | Failed`), and `follow_up_context.py`. After rebasing, the OCI handler needs to implement `check_hosts_status` returning `CheckHostsStatusResult` so `RequestStatusService` can drive its state machine. The `EC2FleetHandler` pattern is the reference.

Main's `providers/registration.py` now has `_REGISTERED_PROVIDERS: list[str] = ["aws"]` with a dynamic importlib dispatch loop. After rebasing, the only change needed is appending `"oci"`. The imperative `register_all_provider_types()` rewrite on this branch conflicts and should be dropped.

`initialize_oci_provider(container)` should call into the five extension-point registries that main provides: `CLISpecRegistry`, `DefaultsLoaderRegistry`, `TemplateExtensionRegistry`, `AuthRegistry`, `FieldMappingRegistry`. See `providers/aws/registration.py` for the pattern. Once `TemplateExtensionRegistry` carries an OCI extension, the `_is_oci_template` helpers and OCI branches in `template_handlers.py` can be removed entirely - those checks belong in the extension, not in shared application code.

`TemplateDTO.provider_config` on main is `Optional[BaseModel]` populated via `TemplateExtensionRegistry`; the AWS-specific field extraction in `from_domain` resolves on rebase.

Blocking subprocess calls from `async def` methods need `asyncio.to_thread` wrapping - noted inline on `_run_oci`.

**OCI-specific issues**

`_is_test_context()` / `_force_live_cli_for_tests` / `_use_mock_mode` put test-framework awareness into the production handler. A stray `PYTEST_CURRENT_TEST` in any outer CI harness silently returns mock OCIDs. The correct seam is an injected `CommandRunner` protocol; tests monkeypatch that injection point.

`_run_oci` joins the full command list including `--from-json <payload>` into error messages. If `metadata` contains `ssh_authorized_keys` or `user_data`, those values appear verbatim in every log line. Redact the `--from-json` value before constructing the error string.

`ProviderType` in `domain/base/provider_interfaces.py` has placeholder members `PROVIDER1` and `Provider2` with inconsistent capitalisation and no `OCI` value. `ProviderOperationEvent.provider_type` is now typed as `ProviderType`, so OCI events cannot be constructed. Remove the placeholders and add `OCI = "oci"`.

`private_key_path` is stored from `input()` without path normalisation or existence check. `nsg_ids` are silently dropped in `_launch_with_cli` with no warning. `_build_launch_cli_payload` is dead code. `describe_resource_instances` returns an empty stub indistinguishable from "not found".

**QA gaps**

`_resolve_availability_domain` fallback branches (subnet-lookup and compartment-fallback paths) have no direct tests. `register_oci_services_with_di` and `register_oci_provider_instance` are untested bootstrap entry points. Partial-success multi-instance create (count > 1, mid-batch failure) leaks already-launched instances with no cleanup. `@staticmethod _is_test_context` mock fragility is noted inline.

No Python OCI SDK is added - subprocess to the `oci` CLI binary means no `[oci]` extras group is needed for now.

src/orb/providers/oci/handlers/oci_compute_handler.py
```diff
@@ -0,0 +1,545 @@
...
+    ) -> None:
+        self._logger = logger
+        self._region = region
+        self._profile = profile
+        self._credential_source = credential_source
+        self._oci_cli_available = shutil.which("oci") is not None
+        self._force_live_cli_for_tests = False
+
+    @staticmethod
+    def _is_test_context() -> bool:
```
fgogolli, line 38: `_is_test_context()` reading `PYTEST_CURRENT_TEST`, `_force_live_cli_for_tests` as a production field, and `_use_mock_mode` treating any test env var as a silent live-call suppressor are test-framework concerns embedded in production code. A stray `PYTEST_CURRENT_TEST` set by any outer CI harness returns fake OCIDs from a production deployment. The correct seam is an injected `CommandRunner` protocol (or callable): the handler takes `run_command=subprocess.run` in `__init__`, tests pass a mock callable. Production code stays unaware of pytest.

src/orb/providers/oci/handlers/oci_compute_handler.py
```diff
@@ -0,0 +1,545 @@
...
+            args.extend(["--metadata", json.dumps(metadata)])
+        tags = launch_payload.get("freeform_tags")
+        if isinstance(tags, dict) and tags:
+            args.extend(["--freeform-tags", json.dumps(tags)])
+
+        nsg_ids = vnic.get("nsg_ids")
+        if isinstance(nsg_ids, list) and nsg_ids:
+            self._logger.info(
+                "OCI launch includes nsg_ids in template; CLI flag handling currently skipped for nsg_ids=%s",
+                nsg_ids,
```
fgogolli, line 220: `nsg_ids` are accepted in templates and mapped in `_build_launch_cli_payload`, but silently dropped here with an INFO log. A user providing `security_group_ids` in a template will get instances without network security group attachment and no indication in the response. Either pass the NSG IDs via `--nsg-ids` (the OCI CLI flag) or return an explicit error; logging at INFO with no field in `ProviderResult.metadata` is not actionable.

src/orb/providers/oci/handlers/oci_compute_handler.py
```diff
@@ -0,0 +1,545 @@
...
+                "OCI template requests preemptible capacity; launching with OCI preemptible config"
+            )
+        if force_ondemand and requested_capacity_type == "preemptible":
+            self._logger.info("Retrying OCI launch as on-demand after preemptible attempt failure")
+
+        return self._run_oci(args, region_override=inferred_region)
+
+    @staticmethod
+    def _build_launch_cli_payload(launch_payload: dict[str, Any]) -> dict[str, Any]:
+        vnic = launch_payload.get("create_vnic_details", {})
```
fgogolli, line 233: `_build_launch_cli_payload` is never called from `_launch_with_cli` or anywhere else in the class - `_launch_with_cli` constructs CLI args directly. This appears to be a leftover from an earlier `--from-json` based design. If it's not wired in, remove it; dead untested code that mirrors live logic in a slightly different shape will diverge silently.

src/orb/providers/oci/handlers/oci_compute_handler.py
```diff
@@ -0,0 +1,545 @@
...
+                {"operation": "get_instance_status", "provider": "oci"},
+            )
+
+        return ProviderResult.success_result(
+            {"instances": statuses},
+            {"operation": "get_instance_status", "provider": "oci"},
+        )
+
+    async def describe_resource_instances(self, operation: ProviderOperation) -> ProviderResult:
+        params = operation.parameters or {}
```
fgogolli, line 524: `describe_resource_instances` returns `instances: []` for every ID with `success=True` and no CLI call. A caller polling post-provisioning status gets a successful empty response it cannot distinguish from "instance not found". Either implement the OCI CLI lookup (`oci compute instance get --instance-id <id>`) or return a `NOT_IMPLEMENTED` error code so callers can handle the gap explicitly.

src/orb/application/services/orchestration/create_template.py
```diff
@@ -2,36 +2,97 @@
...
 from orb.application.ports.query_bus_port import QueryBusPort
 from orb.application.services.orchestration.base import OrchestratorBase
 from orb.application.services.orchestration.dtos import CreateTemplateInput, CreateTemplateOutput
 from orb.application.template.commands import CreateTemplateCommand
 from orb.domain.base.ports.logging_port import LoggingPort
+from orb.domain.template.ports.template_defaults_port import TemplateDefaultsPort
+
+
+def _is_oci_template(configuration: dict, provider_name: Optional[str]) -> bool:
+    provider_type = configuration.get("provider_type") or configuration.get("providerType")
```
fgogolli, line 17: `_is_oci_template` is a provider-specific predicate in a provider-agnostic orchestrator - a direct violation of the extension-point design that `TemplateExtensionRegistry` (on main) exists to prevent. After rebasing and registering an `OCITemplateExtension`, this function and the guard around `_template_defaults_service` can be removed. The guard should be `self._template_defaults_service is not None` only; the service already knows which providers it handles. The same function is duplicated verbatim in `validate_template.py` and `template_handlers.py`.

src/orb/application/services/orchestration/validate_template.py
```diff
@@ -11,27 +13,66 @@
     ValidateTemplateOutput,
 )
 from orb.domain.base.ports.logging_port import LoggingPort
+from orb.domain.template.ports.template_defaults_port import TemplateDefaultsPort
+
+
+def _is_oci_template(configuration: dict, provider_name: Optional[str]) -> bool:
+    provider_type = configuration.get("provider_type") or configuration.get("providerType")
```
fgogolli, line 20: Same `_is_oci_template` duplication as `create_template.py`. A bug fix in one site silently diverges from the other two. After registering an OCI template extension on rebase, this guard and the injected `TemplateDefaultsPort` can be removed here too.

src/orb/application/commands/template_handlers.py
```diff
@@ -95,6 +95,23 @@ async def execute_command(self, command: CreateTemplateCommand) -> None:
             "image_id": command.image_id,
             "tags": command.tags,
         }
+        provider_api_name = str(dto_fields.get("provider_api") or "").lower()
+        if provider_api_name.startswith("oci"):
+            metadata = dict(dto_fields.get("metadata") or {})
```
fgogolli, line 100: OCI-specific metadata-merging logic directly inside a shared CQRS command handler. `FieldMappingRegistry` and `TemplateExtensionRegistry` on main exist to keep this kind of provider-specific field normalisation out of shared handlers. After registering an `OCITemplateExtension.normalise_for_storage(dto_fields)` method, this branch becomes `extension.normalise_for_storage(dto_fields)` if an extension is registered - no provider-type string check in shared code.

src/orb/providers/oci/handlers/oci_compute_handler.py
```diff
@@ -0,0 +1,545 @@
...
+                f"STDOUT: {completed.stdout.strip()}\nSTDERR: {completed.stderr.strip()}"
+            )
+
+        raw = completed.stdout.strip()
+        if not raw:
+            return {}
+        return json.loads(raw)
+
+    def _resolve_availability_domain(
+        self,
```
fgogolli, line 99: `_resolve_availability_domain` has three non-trivial branches: subnet-OCID lookup, compartment fallback list, and the `None` return path. None of these are tested directly - all test paths pass `availability_domain` explicitly, bypassing this method. The subnet-lookup and compartment-fallback branches need focused tests, including the case where both fail and `None` is returned to the caller.

fgogolli (commented):
Supplemental inline comments on items not anchored in the original review. Anchors: `_is_test_context` production anti-pattern, `nsg_ids` silently dropped in the live CLI path, dead `_build_launch_cli_payload`, hollow `describe_resource_instances` stub, the three duplicated `_is_oci_template` sites in shared application code, and the untested `_resolve_availability_domain` fallback branches.

fgogolli:
Now that #285 has merged, this PR needs a rebase with a few adjustments.

`providers/registration.py` on main uses `_REGISTERED_PROVIDERS = ["aws"]`. Please add `"oci"` to that list instead of calling `register_oci_provider` directly.

Correctness item: `OCIProviderStrategy` currently declares `get_available_credential_sources`, `test_credentials`, `get_credential_requirements`, and `get_operational_requirements` as instance methods. `init_command_handler` invokes them on the class before any strategy instance exists — please decorate them with `@classmethod`. Also add `get_cli_provider_config` as `@classmethod` so `orb init --provider-type oci` populates the provider config block correctly.

`_interactive_setup`, `_configure_additional_provider`, and `_write_config_file` in `init_command_handler.py` were substantially rewritten in #285 to remove `region`/`profile` positional coupling and use an opaque `provider_config` dict. The OCI-specific helper additions (`_collect_credential_source_inputs`, `_copy_provider_auth_config`, `_credential_test_kwargs`) will need to be grafted onto the current function bodies.

Finally, please verify whether `_preload_explicit_config` in `cli/main.py` is still needed given the `Application(config_path)` fix that shipped in #285 — that fix ensures an explicit `config_path` is always honoured, which may make the preload redundant.

OCI provider files themselves are self-contained and should apply cleanly.

fgogolli:

- `CLISpecRegistry`
- `FieldMappingRegistry`
- `DefaultsLoaderRegistry`
- `TemplateExtensionRegistry`

PR #253, opened 2026-06-15
feat: provider extension points, ProviderFulfilment contract, auth security fixes

fgogolli:
## Description

Adds provider extension points so new providers (Azure, GCP, OCI, etc.) can be added cleanly without leaking provider-specific knowledge into shared services. Scope grew during review: the branch now also includes an auth subsystem refactor with typed sub-configs, four security fixes found during an auth audit, AWS error detail surfacing, provider registration consolidation, a **ProviderFulfilment domain contract** that fixes weighted-fleet status semantics, an **MCP server bootstrap fix**, several live-test scheduler/wire-format corrections, and CI/test infrastructure improvements.

**Motivation:** In-flight provider PRs (#234 OCI, plus #256/#257/#258 — the Azure + GCP work originally proposed in #240, now split into a common-scaffold PR plus one per provider) were each touching 50+ shared files because `main` lacked proper extension points. This PR adds those extension points and tightens the seams so new provider PRs are small and reviewable.

After this lands, **adding a new provider touches: 1 line in `_REGISTERED_PROVIDERS` + `providers/<name>/` package + SDK deps**.

---

### Extension points added (8)

1. **`ProviderRegistration.default_api`** — provider declares its default API at registration; `ProviderRegistry.get_default_api()` exposes it; `TemplateDefaultsService` delegates instead of reading provider-internal config directly. Removes hardcoded `"aws"` defaults from `api/routers/templates.py`.
2. **`CLISpecRegistry`** enforcement — `cli/args.py`, `init_command_handler.py`, `infrastructure_command_handler.py`, `provider_config_handler.py` all drive provider flags/display/config from the registry. No more hardcoded `--aws-profile`/`--aws-region` lines per consumer.
3. **`TemplateExtensionRegistry`** — `TemplateDTO.provider_config: BaseModel | None` typed field populated via the registry. Replaces the `getattr(template, "fleet_type", None)` string-literal pattern in `from_domain`. AWS-specific fields move into `AWSTemplateDTOConfig`.
4. **`OperationOutcome`** typed discriminated union (`Accepted | Completed | RequiresFollowUp | Failed`). Replaces stringly-typed `provider_metadata` keys. Provider strategies declare actual operation state explicitly; orchestrator pattern-matches with `assert_never()` for exhaustiveness.
[...]

PR #254, opened 2026-06-15
feat: add UI for ORB

parth-pandit:
## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependencies update

## Related Issues

Fixes #

## How Has This Been Tested?

- [ ] Manual testing performed

## Test Configuration
* Python version:
* OS:
* AWS region:
* Dependencies changed:

## Checklist

- [ ] My code follows the project's style guidelines
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes

## Performance Impact

- [ ] No significant performance impact
- [ ] Performance improved

## Security Considerations

- [ ] Security improved

## Reviewers

## Co-authors
- Parth Pandit (@parth-pandit)
- Mohamad Naji 
- Rohan Aditya Ram (@rohanaditya)


fgogolli (commented):
This PR was authored before #253 merged. After rebasing, the diff will shrink to the `ui/` tree only - all 41 changed files are already under `ui/`. The non-`ui/` artefacts visible in the current diff are pre-#253 baseline noise and will drop automatically on rebase.

The real scope is a standalone React 19 + Vite 8 + Tailwind 4 SPA. Component architecture, routing, and state layering (TanStack Query v5 for server state, Context for UI state) are sound. The issues below need resolving before this can merge.

**Phantom backend endpoints**

`ui/src/api/aws.js` calls seven routes that don't exist on main: `/aws/discovery`, `/aws/regions`, `/aws/confirm-shutdown`, `/aws/sync-template`, `/aws/cleanup-template`, `GET /config/defaults`, and `PUT /config/defaults`. `server.py` registers only `machines`, `requests`, and `templates`. Every call into `aws.js` returns 404 at runtime. The `useCreateRequest` hook silently swallows errors from `getUiDefaults` and the post-create `syncAwsTemplate` call in a bare `catch {}`. Either land the backend router before merging, or gate those UI features behind a feature flag and remove the silent swallow.

**No auth integration**

#253 added `AuthMiddleware`. The axios client sends no `Authorization` header and has no interceptor to inject one. The `EventSource` in `requests.js` is constructed without `withCredentials`. With `auth.enabled = true`, every UI call returns 401 and the entire interface goes dark. Add a token-acquisition flow, an axios request interceptor reading from `sessionStorage`, and for SSE - which cannot send headers - use a short-lived query-param token issued by the server.

**Plaintext HTTP fallback and CSRF**

`VITE_ORB_API_BASE` defaults to `http://localhost:8000/api/v1` in three files. A production container built without that env var silently ships cleartext. Fail the Vite build when the variable is unset in production mode. Separately, the axios client sends no CSRF token and `CORSConfig.origins` defaults to `["*"]`; state-changing endpoints are reachable from any origin when auth is also off.

**Accessibility**

All four modals (`ConfirmModal`, `CreateTemplateModal`, `EditTemplateModal`, `RequestModal`) lack `role="dialog"`, `aria-modal="true"`, `aria-labelledby`, and a focus trap. `StatusDot` conveys status by colour only - no text alternative for screen readers or colour-blind users. The toast container has no `aria-live` region. Unicode nav icons in `Sidebar` lack `aria-hidden="true"`, causing screen readers to verbalise glyph names. Filter inputs and checkboxes in `Machines.jsx` have no labels. Adding `eslint-plugin-jsx-a11y` would catch most of these statically.

**Architecture and quality**

`CreateTemplateModal.jsx`, `EditTemplateModal.jsx`, and `RequestModal.jsx` live in `pages/` but are overlays, not routes. `Templates.jsx` and both template modals call `client.delete`/`client.post`/`client.put` directly, bypassing the `api/templates.js` module. The `onerror` handler in `Requests.jsx` calls `es.close()` unconditionally, permanently killing the stream and defeating `EventSource`'s built-in reconnection. There is no React Error Boundary - a render-phase throw in any page produces a blank screen.

`CreateTemplateModal.jsx` is 558 lines with four `eslint-disable` suppressions papering over stale-closure bugs; `EditTemplateModal.jsx` is 497 lines with the same pattern. The `useRequests` query key omits `filter`, so all filter variants share one cache entry. There is no test framework, no test files, and no CI coverage for `ui/` - no Node.js setup, no `npm ci`, no `vite build`, no `npm audit` in any workflow. Add a `ui-ci.yml` running install, lint, build, and audit on every PR touching `ui/**`.

ui/src/api/client.js
```diff
@@ -0,0 +1,25 @@
+import axios from 'axios'
+
+const client = axios.create({
+  baseURL: import.meta.env.VITE_ORB_API_BASE || 'http://localhost:8000/api/v1',
```
fgogolli, line 4: Two issues here. First, the `http://` fallback means any production or staging build where `VITE_ORB_API_BASE` is unset will send all API traffic - including AWS resource IDs and request payloads - in cleartext. Fail the build instead: add a check in `vite.config.js` that throws when `VITE_ORB_API_BASE` is unset and `NODE_ENV === 'production'`. Second, there is no `Authorization` header and no request interceptor to inject one. `AuthMiddleware` (from #253) rejects every call with 401 when auth is enabled. Add an interceptor that reads a token from `sessionStorage` and attaches `Authorization: Bearer <token>` before each request.

ui/src/api/aws.js
```diff
@@ -0,0 +1,57 @@
+import client from './client'
+
+// Returns { region, vpcs, subnets, security_groups, amis }
+export const getAwsDiscovery = async (region = 'us-east-1') => {
+  const { data } = await client.get('/aws/discovery', { params: { region } })
```
fgogolli, line 5: None of these seven endpoints exist on the backend yet - there is no `aws` router registered in `server.py`, and `GET /config/defaults` / `PUT /config/defaults` have no `config` router either. After rebase against main, every call here returns 404. If a backend PR is in flight, note the dependency in the PR description and gate the UI features (the Config page and the AWS discovery flow in `CreateTemplateModal`) behind a feature flag until the router lands. The silent `catch {}` in `useCreateRequest` that swallows `getUiDefaults` and `syncAwsTemplate` errors should at minimum log to console so failures are visible during development.

ui/src/api/requests.js
```diff
@@ -0,0 +1,38 @@
...
+  return data
+}
+
+// Server-Sent Events stream for a request's status. Returns an EventSource.
+export const streamRequest = (id, { interval = 2, timeout = 300 } = {}) => {
+  const base =
+    import.meta.env.VITE_ORB_API_BASE || 'http://localhost:8000/api/v1'
+  const url = new URL(`${base}/requests/${id}/stream`)
+  url.searchParams.set('interval', String(interval))
+  url.searchParams.set('timeout', String(timeout))
```
fgogolli, line 25: Two issues. `EventSource` is constructed without `withCredentials: true` - when the server uses cookie-based session auth the browser won't attach credentials to the SSE connection, and the stream silently fails to authenticate. Add `{ withCredentials: true }` as the second argument. Also, the `onerror` handler in `Requests.jsx` calls `es.close()` unconditionally on every error event, permanently killing the stream and defeating `EventSource`'s built-in reconnection. Only close on terminal errors (where `es.readyState === EventSource.CLOSED` after the event); let transient network drops reconnect automatically.

ui/src/api/machines.js
```diff
@@ -0,0 +1,19 @@
...
+export const getMachine = async (id) => {
+  const { data } = await client.get(`/machines/${id}`)
+  return data
+}
+
+// Both single and bulk use the same endpoint; body is { machineIds: [...] }
+// We always send force=true so a stale "pending return" record from a prior
+// failure (e.g. expired AWS creds) doesn't block subsequent attempts.
+export const returnMachines = async (ids) => {
+  const { data } = await client.post('/machines/return', { machineIds: ids, force: true })
```
fgogolli, line 17: `force: true` is hardcoded with no user-facing warning. The comment explains the intent (clear stale pending-return records), but operators using the UI have no indication that a forced return bypasses the pending-return guard. If this is always the right behaviour then it's fine, but if the guard exists for a reason - e.g. to prevent double-returns on in-flight requests - users need a disclosure before the action completes.

ui/src/components/ConfirmModal.jsx
```diff
@@ -0,0 +1,55 @@
...
+  loading = false,
+}) {
+  useEffect(() => {
+    const onKey = (e) => { if (e.key === 'Escape') onCancel() }
+    window.addEventListener('keydown', onKey)
+    return () => window.removeEventListener('keydown', onKey)
+  }, [onCancel])
+
+  return (
+    <div className="fixed inset-0 z-50 flex items-center justify-center bg-black/40 dark:bg-black/60">
```
fgogolli, line 20: The inner panel `div` needs `role="dialog"`, `aria-modal="true"`, and `aria-labelledby` pointing to an `id` on the `<h2>`. Without these, screen readers don't announce this as a modal and don't constrain the virtual cursor to its contents. The same gap applies to `CreateTemplateModal`, `EditTemplateModal`, and `RequestModal`. Escape-key dismiss is correctly wired here but absent in the three page-level modals. A focus trap is also needed: move focus to the first focusable element on open, constrain `Tab`/`Shift+Tab` within the dialog, and restore focus to the trigger element on close.

ui/src/components/Sidebar.jsx
```diff
@@ -0,0 +1,126 @@
+import { NavLink } from 'react-router-dom'
+import { useSystemHealth, useSystemInfo } from '../hooks'
+import StatusDot from './StatusDot'
+
+const NAV = [
+  { to: '/',          label: 'Dashboard', icon: '▦' },
```
fgogolli, line 6: The Unicode icon `<span>` elements need `aria-hidden="true"`. Without it, screen readers attempt to verbalise the glyph names (`black medium square`, `left half black circle`, etc.) alongside the visible label text. The visible label text is sufficient on its own, so hiding the icons is the right fix. No additional `aria-label` on `NavLink` is needed once the icon is hidden.

ui/vite.config.js
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+import tailwindcss from '@tailwindcss/vite'
+
+export default defineConfig({
+  plugins: [react(), tailwindcss()],
```
fgogolli, line 6: Three additions needed. First, add `server.proxy` to forward `/api` and `/health`/`/metrics`/`/info` to `localhost:8000` during development - without it, CORS must be wide open on the backend for local dev, and `VITE_ORB_API_BASE` becomes a mandatory local setup step that isn't documented anywhere (the README is still the Vite scaffold placeholder). Second, add a build-time check that throws when `VITE_ORB_API_BASE` is unset in production mode, to prevent the cleartext HTTP fallback shipping silently. Third, add `preview.headers` with `Content-Security-Policy`, `X-Frame-Options: DENY`, and `X-Content-Type-Options: nosniff` for staging preview builds.

ui/src/pages/Requests.jsx
```diff
@@ -0,0 +1,377 @@
...
+        if (!e.data || e.data.trim() === '{}') return
+        try {
+          const payload = JSON.parse(e.data)
+          const first = payload?.requests?.[0]
+          if (first) setStreamed(first)
+        } catch {
+          // ignore malformed frames
+        }
+      }
+      es.onerror = () => {
```
fgogolli, line 118: Calling `es.close()` on every `onerror` event permanently kills the stream. `EventSource` reconnects automatically after transient network drops by design - closing it here disables that and falls back silently to the 10-second React Query polling. Only close when the connection is definitively gone: check `es.readyState === EventSource.CLOSED` after the error event, or distinguish HTTP error status codes (e.g. 404, 401) from transient drops before deciding to close.

ui/src/components/StatusDot.jsx
```diff
@@ -0,0 +1,30 @@
...
+  timeout:   'bg-red-500',
+  error:     'bg-red-500',
+  unknown:   'bg-gray-400',
+}
+
+export default function StatusDot({ status, size = 'md' }) {
+  const key = String(status || '').toLowerCase().trim()
+  const color = DOT_STYLES[key] || 'bg-gray-400'
+  const sz = size === 'sm' ? 'w-2 h-2' : 'w-2.5 h-2.5'
+  return <span className={`inline-block rounded-full ${sz} ${color}`} />
```
fgogolli, line 29: Status is conveyed by colour only. Screen readers see an empty inline element; colour-blind users get no textual cue. Add `aria-label={status}` to this `<span>` - or a visually-hidden sibling `<span className="sr-only">{status}</span>` if the dot must remain a standalone element. The `DOT_STYLES` map already normalises the key, so the same value can be used as the label.

ui/src/components/Toast.jsx
```diff
@@ -0,0 +1,60 @@
...
+    if (type === 'success') {
+      setTimeout(() => dismiss(id), duration)
+    }
+    return id
+  }, [dismiss])
+
+  return (
+    <ToastCtx.Provider value={{ show, dismiss }}>
+      {children}
+      <div className="fixed bottom-4 right-4 z-50 flex flex-col gap-2 w-80">
```
fgogolli, line 26: The toast container has no `aria-live` region. Notifications appear visually but are never announced by screen readers. Add `aria-live="polite"` and `aria-atomic="true"` to this `div`. Also, the dismiss button (`×`) at line 47 has no `aria-label` - add `aria-label="Dismiss"` so its purpose is clear to assistive technology.

ui/src/pages/CreateTemplateModal.jsx
```diff
@@ -0,0 +1,558 @@
...
+
+  // Initialise VPC + AMI defaults once discovery returns
+  useEffect(() => {
+    if (defaultVpcId && !vpcId) setVpcId(defaultVpcId)
+  }, [defaultVpcId, vpcId])
+
+  useEffect(() => {
+    if (!form.imageId && amis[0]) update('imageId', amis[0].id)
+  }, [amis]) // eslint-disable-line react-hooks/exhaustive-deps
+
```
fgogolli, line 160: Four `// eslint-disable-line react-hooks/exhaustive-deps` suppressions in this file (lines 160, 175, 179, and a similar pattern in `EditTemplateModal`) are masking stale-closure bugs rather than fixing them. The `update` callback is recreated each render but omitted from deps; the `amis`, `subnetsForVpc`, and `sgsForVpc` effects fire with stale references. Extract the form update logic into a `useTemplateForm` hook with stable `useCallback`-wrapped setters, then include all dependencies correctly. At 558 lines this component is also doing too much - splitting the hook from the render will make both testable.

ui/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "orb-ui",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "vite build",
+    "lint": "eslint .",
+    "preview": "vite preview"
```
fgogolli, line 10: No `test` script and no test runner (`vitest`, `jest`, `@testing-library/react`) in devDependencies. The UI ships with zero automated coverage. Add `vitest` and `@testing-library/react` at minimum; the highest-risk logic to cover first is the `useCreateRequest` mutation (multi-step with silent error swallowing) and the modal components. Also add an `"audit": "npm audit --audit-level=moderate"` script and wire it into CI - the `security-code.yml` workflow currently covers Python only and won't catch Node dependency CVEs. The `@types/react` and `@types/react-dom` devDependencies are unused (no `.ts`/`.tsx` files, no `tsconfig.json`); remove them or commit to TypeScript.

ui/README.md
```diff
@@ -0,0 +1,16 @@
+# React + Vite
```
fgogolli, line 1: This is the unmodified Vite scaffold placeholder. Replace it with ORB UI documentation: how to set `VITE_ORB_API_BASE`, that it is a required variable in production builds, how to run the dev server against a local ORB instance, and the note that all `VITE_*` variables are embedded in the built bundle and must not contain secrets.

fgogolli:
Closing this PR in favour of #284 as we have decided to use a framework that allows us to embed the UI into the PyPi package for the project and be delivered as one single binary. Thanks very much for your work and contribution though, as the UI created during this PR/Hackathon was used as a vision to frame the new Reflex UI and the ideas have been retained. Thanks again and looking forward to your future contributiouns!

PR #241, opened 2026-06-09
fix(storage): repair SQL/DynamoDB strategies + add cross-backend contract tests

fgogolli:
## Description

The SQL storage mode was non-functional. Investigation found three independent root causes that each prevented `SQLStorageStrategy` from being constructed or used. DynamoDB had a related batch-operation bug exposed once tests began exercising it. None of these were caught because no tests existed for the strategies themselves — only for low-level components.

This PR fixes all four issues, removes a dead Protocol API that was hiding the SQL bug, and adds a parametrised contract test suite that exercises every storage backend end-to-end.

### Bugs fixed

| # | Backend | Bug |
|---|---------|-----|
| 1 | SQL | `SQLQueryBuilder` extended abstract `QueryManager` but did not implement `build_query` / `execute_query` / `validate_query` → `TypeError: Can't instantiate abstract class` on construction. |
| 2 | SQL | `SQLStorageStrategy.save_batch` passed raw SQL strings to `session.execute(...)`, which SQLAlchemy 2.x rejects without `text(...)` wrapping. |
| 3 | UoW | `UnitOfWork` Protocol declared `register_new` / `register_dirty` / `register_removed` (Fowler-style identity-map UoW) that no concrete UoW implemented and no caller invoked. Pyright complained, and the noise was silenced via `# type: ignore[abstract]`. |
| 4 | DynamoDB | `DynamoDBTransactionManager` built `TransactItems` with plain Python dicts; `transact_write_items` requires AttributeValue-tagged dicts, so every batch put / delete / update was rejected. |

### What changed

- `SQLQueryBuilder` — added `build_query`, `execute_query`, `validate_query` implementations satisfying `QueryManager`
- `SQLStorageStrategy.save_batch` — wrapped query in `text(...)` for SQLAlchemy 2.x
- `UnitOfWork` Protocol — removed unused `register_new` / `register_dirty` / `register_removed`. The codebase uses direct repository writes (`uow.machines.save(...)`), not the identity-map pattern. Re-add the methods only if and when that pattern is actually adopted.
- `DynamoDBTransactionManager` — added `boto3.dynamodb.types.TypeSerializer` at the Put / Delete / Update boundary
- Removed three `# type: ignore[abstract]` suppressions that hid the SQL abstract issue

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[...]

PR #255, opened 2026-06-15
fix: provider storage strategy 

kirillsc:
## Description

The DynamoDB storage backend was non-functional out of the box. Standing up the create → read → write → read-back loop hit a separate defect at each consecutive stage, each masking the next. This PR fixes the four blockers, adds the round-trip test coverage that would have caught them, and registers the Aurora backend that was implemented but never wired in.

**What:**
- Validator derives valid `storage.strategy` values from the storage registry instead of a hardcoded `["json", "sql"]`, so a registered backend (e.g. `dynamodb`, `aurora`) is accepted while the generic schema stays provider-agnostic.
- `AWSClient` gains lazy `dynamodb_client`/`dynamodb_resource` accessors (mirroring `ec2_client`); the DynamoDB UoW factory stops handing a raw boto3 client to wrapper-expecting code.
- DynamoDB converter: serialise `bool` before `int` (bool is an int subclass → was crashing on `Decimal("True")`); stop pre-parsing ISO timestamps to `datetime` on read (the repository layer owns parsing, matching the JSON backend).
- Register the existing Aurora backend in `register_all_storage_types()`.

**Why:** Each defect gates one stage of the only meaningful DynamoDB workflow (init → create → status), so the backend cannot provision or read back a single instance without all four fixes. Existing tests only round-tripped `{id, name}`, so the bool/timestamp regressions were never exercised.

**How:** Minimal, in-idiom changes scoped to the storage layer, preserving the provider-agnostic separation of the generic config layer (a prior refactor moved AWS storage code under `providers/aws/`). Backend selection is registry-driven; the converter fixes match the JSON backend's proven behaviour.

**Testing:** New converter unit tests + moto-backed contract/round-trip tests; reverting either converter fix turns the new integration tests red. Exercised end-to-end against real AWS (`eu-west-1`).

## Type of Change

## Related Issues
Fixes #

## How Has This Been Tested?

New tests: `test_dynamodb_converter.py` (converter unit), `TestDynamoDBRichTypes` in `test_storage_contract.py` (bool/number/timestamp round-trip via moto), `test_request_roundtrip_dynamodb.py` (Request save → get_by_id), registry-driven validator tests in `test_aws_storage_config.py`. 77 storage tests + 2300+ architecture/config tests pass locally; storage-leak detection green; ruff clean. Manually ran create → status → terminate via `RunInstances` against real AWS.

## Test Configuration
* Python version: 3.13
* OS: macOS (darwin)
* AWS region: eu-west-1
* Dependencies changed: none

## Checklist

## Additional Notes

Follow-ups intentionally **not** in this PR:
- `region`/`profile` on `DynamodbStrategyConfig` duplicate `AWSProviderConfig.region/.profile` and should be removed so storage config carries only `table_prefix`; parked pending a multi-provider resolution rule (which provider supplies the storage block when several are configured).
- Aurora is registered and contract-tested via sqlite; live Aurora MySQL connectivity is not yet covered.

## Performance Impact

## Security Considerations

## Dependencies

None.

## Migration Guide

Not a breaking change. To use DynamoDB, set top-level `storage.strategy: "dynamodb"` and provide `provider.providers[N].config.storage.dynamodb` (`table_prefix`, plus provider `region`/`profile`).

## Deployment Notes

DynamoDB tables are auto-created on first use (partition key `id`, on-demand billing). Requires `dynamodb:CreateTable`/`DescribeTable` plus read/write permissions.

## Reviewers

### DIFF
--- a/src/orb/config/schemas/storage_schema.py
+++ b/src/orb/config/schemas/storage_schema.py
@@ -122,6 +122,18 @@ class RetryConfig(BaseModel):
         return v
 
 
+def _get_valid_storage_strategies() -> set[str]:
+    """Accepted ``storage.strategy`` values: generic baseline + registered backends."""
+    valid = {"json", "sql"}
+    try:
+        from orb.infrastructure.storage.registry import get_storage_registry
+
+        valid.update(get_storage_registry().get_registered_storage_types())
+    except Exception:
+        pass  # registry not ready -> baseline only
+    return valid
+
+
 class StorageConfig(BaseModel):
     """Storage configuration."""
 
@@ -133,10 +145,10 @@ class StorageConfig(BaseModel):
     @field_validator("strategy")
     @classmethod
     def validate_strategy(cls, v: str) -> str:
-        """Validate storage strategy."""
-        valid_strategies = ["json", "sql"]
+        """Validate storage strategy against registered backends (provider-agnostic)."""
+        valid_strategies = _get_valid_storage_strategies()
         if v not in valid_strategies:
-            raise ValueError(f"Storage strategy must be one of {valid_strategies}")
+            raise ValueError(f"Storage strategy must be one of {sorted(valid_strategies)}")
         return v
 
     @model_validator(mode="after")
--- a/src/orb/config/schemas/storage_schema.py
+++ b/src/orb/config/schemas/storage_schema.py
@@ -130,7 +130,10 @@ def _get_valid_storage_strategies() -> set[str]:
 
         valid.update(get_storage_registry().get_registered_storage_types())
     except Exception:
-        pass  # registry not ready -> baseline only
+        # Validator may run before bootstrap registers backends, and storage_schema
+        # must stay importable without orb.infrastructure.storage being initialised.
+        # Fall back to the baseline so config validation never depends on registry health.
+        pass
     return valid
 
 
--- a/src/orb/infrastructure/storage/registration.py
+++ b/src/orb/infrastructure/storage/registration.py
@@ -22,9 +22,13 @@ def register_all_storage_types() -> None:
 
     register_sql_storage()
 
-    from orb.providers.aws.storage.registration import register_dynamodb_storage
+    from orb.providers.aws.storage.registration import (
+        register_aurora_storage,
+        register_dynamodb_storage,
+    )
 
     register_dynamodb_storage()
+    register_aurora_storage()
 
 
 def get_available_storage_types() -> list:
@@ -57,6 +61,14 @@ def get_available_storage_types() -> list:
         pass
 
         available_types.append("dynamodb")
+    except ImportError:
+        pass
+
+    # Check Aurora storage availability
+    try:
+        pass
+
+        available_types.append("aurora")
     except ImportError:
         pass
 
--- a/src/orb/infrastructure/storage/registration.py
+++ b/src/orb/infrastructure/storage/registration.py
@@ -38,39 +38,19 @@ def get_available_storage_types() -> list:
     Returns:
         List of storage type names that are available for registration
     """
+    from importlib.util import find_spec
+
     available_types = []
 
-    # Check JSON storage availability
-    try:
-        pass
-
+    # Each backend is available when its registration module can be imported.
+    # find_spec probes importability without binding an unused name.
+    if find_spec("orb.infrastructure.storage.json.registration") is not None:
         available_types.append("json")
-    except ImportError:
-        pass
-
-    # Check SQL storage availability
-    try:
-        pass
-
+    if find_spec("orb.infrastructure.storage.sql.registration") is not None:
         available_types.append("sql")
-    except ImportError:
-        pass
-
-    # Check DynamoDB storage availability
-    try:
-        pass
-
+    if find_spec("orb.providers.aws.storage.registration") is not None:
         available_types.append("dynamodb")
-    except ImportError:
-        pass
-
-    # Check Aurora storage availability
-    try:
-        pass
-
         available_types.append("aurora")
-    except ImportError:
-        pass
 
     return available_types
 
--- a/src/orb/providers/aws/storage/registration.py
+++ b/src/orb/providers/aws/storage/registration.py
@@ -74,12 +74,6 @@ def create_dynamodb_unit_of_work(config: Any) -> Any:
     Returns:
         DynamoDBUnitOfWork instance with correctly configured AWS client
     """
-    from botocore.config import Config
-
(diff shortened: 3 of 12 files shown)


src/orb/config/schemas/storage_schema.py
```diff
@@ -122,6 +122,18 @@ def validate_max_attempts(cls, v: int) -> int:
         return v
 
 
+def _get_valid_storage_strategies() -> set[str]:
+    """Accepted ``storage.strategy`` values: generic baseline + registered backends."""
+    valid = {"json", "sql"}
+    try:
+        from orb.infrastructure.storage.registry import get_storage_registry
+
```
fgogolli, line 130: **nit:** bare `except Exception: pass` swallows more than just "registry not ready" — import errors and misconfigured registry instances are silently absorbed too.

Schema must work pre-bootstrap, so the fallback itself is correct, but the comment understates what's caught. Suggest expanding so the trade-off is legible:

```python
except Exception:
    # Validator may run before bootstrap registers backends, and storage_schema
    # must stay importable without orb.infrastructure.storage being initialised.
    # Fall back to the baseline so config validation never depends on registry health.
    pass
```

Non-blocking.

src/orb/providers/aws/storage/registration.py
```diff
@@ -90,15 +84,15 @@ def create_dynamodb_unit_of_work(config: Any) -> Any:
...
 
-        session = AWSSessionFactory.create_session(
-            profile=dynamodb_config.profile if dynamodb_config.profile else None,
-            region=dynamodb_config.region,
-        )
-        aws_client = session.client("dynamodb", config=boto_config)
+        # ORB AWS wrapper exposes dynamodb_client/dynamodb_resource (lazy boto3),
+        # which DynamoDBClientManager consumes directly. ConfigurationManager
+        # provides the configuration access AWSClient needs at runtime.
+        aws_client = AWSClient(config=config, logger=_logger)  # type: ignore[arg-type]
```
fgogolli, line 95: **nit:** `# type: ignore[arg-type]` is silencing a real type mismatch — `ConfigurationManager` doesn't satisfy whatever protocol `AWSClient(config=...)` expects.

OK to land here, but worth a follow-up: tighten `AWSClient`'s ctor signature to accept `ConfigurationManager` (or its port), or adapt at this call site. Leaving the ignore long-term means pyright won't catch a real signature regression.

src/orb/providers/aws/storage/registration.py
```diff
@@ -115,18 +109,14 @@ def create_dynamodb_unit_of_work(config: Any) -> Any:
...
         region = config.get("region") or None
         profile = config.get("profile")
         table_prefix = config.get("table_prefix", "hostfactory")
 
-        session = AWSSessionFactory.create_session(
-            profile=profile if profile else None, region=region
-        )
-        aws_client = session.client("dynamodb", config=boto_config)
-
         return DynamoDBUnitOfWork(  # type: ignore[abstract]
```
fgogolli: **nit:** the two `# type: ignore[abstract]` on `DynamoDBUnitOfWork(...)` (this and L102) should now be unnecessary. PR #241 dropped the unused `register_new`/`register_dirty`/`register_removed` from the `UnitOfWork` protocol, so `BaseUnitOfWork` no longer leaks abstract methods.

Worth verifying pyright stays clean after removing both. Same pattern at `registration.py:276` for `SQLUnitOfWork`. Out of scope here but a quick follow-up.

tests/unit/providers/aws/storage/test_dynamodb_converter.py
```diff
@@ -0,0 +1,114 @@
...
+def converter() -> DynamoDBConverter:
+    return DynamoDBConverter(partition_key="id")
+
+
+# --- booleans must not be coerced to Decimal --------------------------------
+
+
+def test_bool_true_serialized_as_bool(converter):
+    assert converter._convert_to_dynamodb_type(True) is True
+
```
fgogolli, line 36: **nit:** these tests call `_convert_to_dynamodb_type` / `_convert_from_dynamodb_type` directly (private API). Couples the test to internal method names — a rename of either private helper will break tests even when behaviour is unchanged.

The public-API round-trip is already covered by `test_request_item_roundtrip_preserves_types` at the bottom, which is the right shape. The bool/timestamp scalar cases could be inlined into single-key dicts to the same effect:

```python
item = converter.to_dynamodb_item("x", {"id": "x", "flag": True})
back = converter.from_dynamodb_item(item)
assert back["flag"] is True
```

Non-blocking — the current tests do catch the regressions.

src/orb/infrastructure/storage/registration.py
```diff
@@ -34,31 +38,19 @@ def get_available_storage_types() -> list:
...
+    from importlib.util import find_spec
 
-    # Check JSON storage availability
-    try:
-        pass
+    available_types = []
 
+    # Each backend is available when its registration module can be imported.
+    # find_spec probes importability without binding an unused name.
+    if find_spec("orb.infrastructure.storage.json.registration") is not None:
```
fgogolli, line 47: Nice cleanup — `find_spec` is the right primitive here. The old `try: pass / except ImportError: pass` blocks were no-ops (`pass` inside `try` can't raise), so they were effectively claiming every backend was always available. Good catch.

fgogolli (commented):
Solid fix. Each bug has its own test; the converter reorder is a textbook `bool ⊂ int` trap, and the timestamp pre-parse vs domain `fromisoformat` is exactly the kind of regression a real-aggregate round-trip would catch. CI is green across pyright/ruff/architecture/security. A few non-blocking nits inline.

fgogolli (dismissed):
Discussed with @kirillsc and will address the broader comments in a future PR. Also need to consider calling registration code for provider-specific storage strategies from outside the per provider codebase. Will ship as is for now.

tests/unit/providers/aws/storage/test_dynamodb_converter.py
kirillsc: ack, fixed

src/orb/providers/aws/storage/registration.py
kirillsc: fixed

src/orb/providers/aws/storage/registration.py
kirillsc: noted

src/orb/config/schemas/storage_schema.py
kirillsc: fixed

fgogolli (approved):

PR #257, opened 2026-06-19
feat: Add GCP provider support

IkeM-L:
The GCP portion of: https://github.com/finos/open-resource-broker/pull/240


fgogolli (commented):
[duplicate test review - please disregard, see full review below]

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
(disregard - duplicate review wrapper)

fgogolli (commented):
This PR was cut before #253 merged. The dominant action item is a rebase onto current main (`dd4c3903a`), which will automatically restore the typed domain contracts the branch is missing. The review below distinguishes between what the rebase resolves automatically and what needs active work on top of it.

**4-PR plan**

The scope needs splitting before merge:

1. **#256 (slimmed)** - only genuinely provider-neutral scaffold that neither AWS nor GCP can own alone.
2. **New AWS-only PR** - `SpotPlacementPlanner`, `SpotPlacementExecutionService`, `AWSSpotPlacementScoreAdapter`, and the `aws_provider_strategy.py` refactor for the `spotPlacementScore` allocation branch. These are EC2 Fleet API mechanics and have no GCP call path.
3. **#257 (this PR, GCP-only)** - After removing the AWS items and Azure vocabulary from the base domain, this PR should contain only `src/orb/providers/gcp/**`, the `google-cloud-compute>=1.39.0` dependency, and calls into the registries that main already provides.
4. **#258** - Azure-specific items (`vm_size`, `vm_sizes`, `placement_*` fields on the base `Template`) belong with the Azure provider work.

**Contracts main now provides**

After rebasing, main (`dd4c3903a`) provides these contracts that the GCP code needs to integrate with:

- `CheckHostsStatusResult(instances, fulfilment=ProviderFulfilment(state, ...))` - the typed verdict returned by `check_hosts_status`. Both `GCPManagedInstanceGroupHandler` and `GCPSingleVMHandler` need to implement this. For MIG the verdict is: compare running instance count against `target_size` to produce `SATISFIED`, `PARTIAL`, or `PENDING`. For Single VM it is simpler: one instance, `RUNNING` maps to `SATISFIED`. Main calls this via `await asyncio.to_thread(handler.check_hosts_status, request)` - please pull that `to_thread` wrapper through in the strategy dispatch.
- `OperationOutcome` discriminated union (`Accepted | Completed | RequiresFollowUp | Failed`) for typed operation results in `provisioning_orchestration_service.py`.
- `provider_config: Optional[BaseModel]` on `TemplateDTO`, populated via `TemplateExtensionRegistryAdapter`. The PR's `dict[str, Any]` version is an older shape; after rebase the typed version takes precedence.
- `DefaultsLoaderRegistry` dispatch in `config/loader.py` - main uses the registry extension point; the PR's hardcoded `{"aws": ..., "gcp": ...}` dict bypasses it and will need updating to call `DefaultsLoaderRegistry.register("gcp", ...)` from `register_gcp_extensions`.

**GCP-specific items to address post-rebase**

These are not rebase conflicts - they are issues with the GCP implementation itself:

- `registration.py` is missing `FieldMappingRegistry.register("gcp", ...)` and `DefaultsLoaderRegistry.register("gcp", ...)` calls that the AWS reference implementation provides. Without them, GCP template queries and field mapping fall back to stubs or raise `KeyError`.
- The module-level `with suppress(Exception): register_gcp_extensions()` block at the bottom of `registration.py` runs at import time and silently discards failures. Call `initialize_gcp_provider(template_factory, logger)` from the DI bootstrap layer instead, matching the AWS pattern.
- `DEFAULT_GCP_SERVICE_ACCOUNT_SCOPES` defaults to `cloud-platform`, which grants full GCP API access. The minimum required scope for Compute Engine is `https://www.googleapis.com/auth/compute`. Please change the default and document the override path.
- `GCPTemplate.service_account_scopes` accepts arbitrary strings with no format validation. Add a `@field_validator` requiring the `https://www.googleapis.com/auth/` prefix.
- `ProviderType` enum in `provider_interfaces.py` has `PROVIDER1 = "provider1"` and `Provider2 = "provider2"` placeholder members with inconsistent casing. Remove them before merge.

**QA gaps**

The zonal MIG code path (`GCPMIGScope.ZONAL`) is entirely untested - `_ComputeClientStub` has no `create_zonal_mig` or `list_zonal_managed_instances` methods, so any test exercising that branch would fail with `AttributeError`. Please add zonal MIG acquire/terminate/status tests before this merges.

`configuration/validator.py` catches all exceptions in `validate_gcp_template` and returns `{"valid": False}`, which will mask programming errors (e.g. `TypeError` from malformed input) as validation failures.

src/orb/domain/base/value_objects.py
```diff
@@ -298,3 +298,21 @@ class PriceType(str, Enum):
     SPOT = "spot"
     RESERVED = "reserved"
     HETEROGENEOUS = "heterogeneous"  # Mix of different pricing types
+
+
+class AllocationStrategy(str, Enum):
+    """Allocation strategy enumeration."""
```
fgogolli, line 304: `AllocationStrategy` and `PlacementSplitStrategy` are EC2 Fleet and Spot Fleet vocabulary - `lowestPrice`, `capacityOptimized`, `spotPlacementScore`, `GREEDY`, `HYBRID` are meaningless outside AWS. After rebase, please move both enums to the AWS-only PR. `AWSAllocationStrategy` already exists in `orb.providers.aws.value_objects`; `PlacementSplitStrategy` belongs there too since its only consumer is `SpotPlacementPlanner`. No GCP code path references either enum.

src/orb/domain/template/template_aggregate.py
```diff
@@ -22,6 +24,8 @@ class Template(BaseModel):
 
     # Instance configuration
     instance_type: Optional[str] = None
+    vm_size: Optional[str] = None
+    vm_sizes: list[str] = Field(default_factory=list)
```
fgogolli, line 28: `vm_size`, `vm_sizes`, and the `placement_*` fields should not appear on the base `Template`. `vm_size`/`vm_sizes` are Azure naming - every Azure consumer is in `providers/azure/`, and GCP uses `instance_type` inherited from `Template` already. The placement fields are wired to `PlacementSplitStrategy` (AWS spot-placement mechanics). After rebase, please move these to the Azure provider PR (#258) and the AWS-only PR respectively. `GCPTemplate` should not inherit fields it does not use.

src/orb/providers/gcp/infrastructure/handlers/mig_handler.py
```diff
@@ -0,0 +1,549 @@
...
+
+        return GCPMutationOutcome(
+            attempted_ids=mig_names,
+            successful_ids=[],
+            operations=operations,
+            warning=self._DELETE_SUBMITTED_WARNING,
+        )
+
+    def check_hosts_status(
+        self,
```
fgogolli, line 233: After rebase, main expects `check_hosts_status` to return `CheckHostsStatusResult(instances, fulfilment=ProviderFulfilment(state, ...))`, called via `await asyncio.to_thread(handler.check_hosts_status, request)` in the strategy. This contract (`provider_fulfilment.py`) was introduced after the branch was cut. For MIG, please implement the verdict by comparing the count of `RUNNING` instances against the MIG `target_size` to produce `SATISFIED`, `PARTIAL`, or `PENDING`. Without this, `RequestStatusService` cannot apply uniform status logic to GCP requests.

src/orb/providers/gcp/registration.py
```diff
@@ -0,0 +1,215 @@
...
+        register_gcp_template_factory(template_factory, logger)
+
+
+def is_gcp_provider_registered() -> bool:
+    """Return whether GCP extensions are registered."""
+    return TemplateExtensionRegistry.has_extension("gcp")
+
+
+with suppress(Exception):
+    register_gcp_extensions()
```
fgogolli, line 214: Module-level side-effects execute at import time. Any transitive import of `orb.providers.gcp` mutates `TemplateExtensionRegistry` and `CLISpecRegistry`, with failures silently discarded by `suppress(Exception)`. `initialize_gcp_provider(template_factory, logger)` is defined 12 lines above and does the same work - call that from the DI bootstrap (`bootstrap/infrastructure_services.py`) instead, alongside the AWS equivalent, so failures surface at startup and test isolation is not undermined by incidental imports.

src/orb/providers/gcp/infrastructure/handlers/base_handler.py
```diff
@@ -0,0 +1,196 @@
...
+    def _operation_wait_timeout_seconds(self) -> int:
+        """Return a bounded wait time for long-running GCP operations."""
+        per_attempt_timeout = self._config.connect_timeout + self._config.read_timeout
+        retry_budget = max(1, self._config.max_retries)
+        return max(1, per_attempt_timeout * retry_budget)
+
+
+def _is_gcp_resource_reference(value: str) -> bool:
+    """Return true when the value is already a GCP resource path or self-link."""
+    return value.startswith(("global/", "regions/", "projects/", "https://", "http://"))
```
fgogolli: Drop `"http://"` from the prefix tuple. GCP Compute Engine self-links are always `https://www.googleapis.com/compute/v1/projects/...`. Accepting `http://` serves no valid use case and would pass a plaintext URL through to the Compute Engine insert request as a network or subnetwork reference without normalisation.

src/orb/providers/gcp/constants.py
```diff
@@ -0,0 +1,8 @@
+"""Shared constants for the GCP provider."""
+
+from __future__ import annotations
+
+
+DEFAULT_GCP_SERVICE_ACCOUNT_SCOPES: tuple[str, ...] = (
+    "https://www.googleapis.com/auth/cloud-platform",
+)
```
fgogolli, line 8: `https://www.googleapis.com/auth/cloud-platform` is the widest possible GCP OAuth scope - it grants full API access for the attached service account. Compute Engine workloads need only `https://www.googleapis.com/auth/compute`. Change the default to `compute` and document that operators must explicitly opt into `cloud-platform` if their workloads require it. Any VM launched without an explicit `service_account_scopes` override currently runs with over-permissioned token access.

src/orb/providers/gcp/domain/template/gcp_template_aggregate.py
```diff
@@ -0,0 +1,103 @@
...
+    project_id: GCPProjectId
+    region: GCPRegion
+    zones: list[GCPZone] = Field(default_factory=list)
+    mig_scope: GCPMIGScope = GCPMIGScope.REGIONAL
+    instance_type: str = Field(default=..., description="Compute Engine machine type")
+    network: Optional[str] = None
+    subnetwork: Optional[str] = None
+    service_account_email: Optional[str] = None
+    service_account_scopes: list[str] = Field(
+        default_factory=lambda: list(DEFAULT_GCP_SERVICE_ACCOUNT_SCOPES)
```
fgogolli, line 42: `service_account_scopes: list[str]` has no validation. Each scope is passed directly to `compute_v1.ServiceAccount(scopes=...)` in `base_handler.py`. Add a `field_validator` asserting each scope matches `^https://www\.googleapis\.com/auth/[\w.]+$`. This blocks malformed values and non-`googleapis.com` schemes from reaching the Compute Engine API silently.

pyproject.toml
```diff
@@ -55,6 +55,7 @@ dependencies = [
     "requests>=2.31.0",
     "PyYAML>=6.0.0", # Exported API (lazy-loaded)
     "jsonschema>=4.17.0",
+    "google-cloud-compute>=1.39.0",
 ]
```
fgogolli, line 59: `google-cloud-compute` is in core dependencies, not an optional extra. Since `uv.lock` pins the resolved version and the library follows semantic versioning, the `>=1.39.0` lower bound is sufficient for day-to-day safety - but please move `google-cloud-compute>=1.39.0` to an optional `[gcp]` extras group (the analysis notes this group is currently commented out in `pyproject.toml`). Keeping a heavy GCP SDK in the core install means all non-GCP deployments pay the import cost.

src/orb/providers/gcp/infrastructure/handlers/single_vm_handler.py
```diff
@@ -0,0 +1,301 @@
...
+            target_ids=target_ids,
+            operation_name="terminate_instance",
+            mutation=lambda instance_name: self._compute_client.delete_instance(
+                zone=zone,
+                instance_name=instance_name,
+            ),
+        )
+
+    def check_hosts_status(
+        self,
```
fgogolli, line 116: After rebase, please implement `check_hosts_status` here to return `CheckHostsStatusResult(instances=..., fulfilment=ProviderFulfilment(state=...))`. Main calls this via `await asyncio.to_thread(handler.check_hosts_status, request)`. For a single VM the verdict is straightforward: one instance, `RUNNING` maps to `SATISFIED`, any transient state maps to `PENDING`. The MIG handler has the same requirement - see the existing inline on `mig_handler.py`.

src/orb/application/services/spot_placement_planner.py
```diff
@@ -0,0 +1,177 @@
+"""Generic spot placement score planning service."""
+
```
fgogolli, line 2: This service exists solely to support the `spotPlacementScore` AWS allocation strategy (driven by EC2 SpotPlacementScore API). The docstring says provider-agnostic but no GCP code path calls it and the only concrete `SpotPlacementScoreAdapter` implementation is `AWSSpotPlacementScoreAdapter`. Please move this file - along with `spot_placement_execution.py` and `spot_placement_score_adapter.py` - to a new AWS-only PR. It has no business in a GCP provider PR.

src/orb/application/services/spot_placement_execution.py
```diff
@@ -0,0 +1,337 @@
+"""Provider-agnostic execution helpers for spot placement plans."""
+
```
fgogolli, line 2: Move to the AWS-only PR together with `spot_placement_planner.py` and `spot_placement_score_adapter.py`. All three files form a unit that supports `spotPlacementScore` fleet allocation on EC2 Fleet. No GCP import path references any of them.

src/orb/providers/aws/infrastructure/services/spot_placement_score_adapter.py
```diff
@@ -0,0 +1,109 @@
+"""AWS spot placement score adapter."""
+
```
fgogolli, line 2: AWS-only adapter for the EC2 SpotPlacementScore API. Move to the new AWS-only PR alongside `spot_placement_planner.py` and `spot_placement_execution.py`. These three files form a coherent feature and should not ship in a GCP provider PR.

src/orb/domain/base/provider_interfaces.py
```diff
@@ -9,6 +9,7 @@ class ProviderType(str, Enum):
     """Supported provider types."""
 
     AWS = "aws"
+    GCP = "gcp"
     PROVIDER1 = "provider1"
     Provider2 = "provider2"
```
fgogolli, line 14: `PROVIDER1 = "provider1"` and `Provider2 = "provider2"` are scaffolding placeholders - `Provider2` also violates `SCREAMING_SNAKE_CASE` enum convention. Neither value is referenced anywhere in the codebase. Please remove both before merge. A closed-world `ProviderType` enum in base domain also creates an OCP issue: adding a third provider later requires a domain edit. The existing string-keyed registries (`"aws"`, `"gcp"`) are the correct extension point.

src/orb/config/loader.py
```diff
@@ -207,6 +200,44 @@ def _load_strategy_defaults(cls, config_manager=None) -> dict[str, Any]:
             get_config_logger().warning("Failed to load scheduler defaults: %s", e)
         return merged
 
+    @classmethod
+    def _load_provider_defaults(cls) -> dict[str, Any]:
+        """Load static provider defaults without bootstrapping provider registries."""
+        merged: dict[str, Any] = {}
+        provider_default_loaders = {
+            "aws": cls._load_aws_provider_defaults,
```
fgogolli, line 208: This hardcoded `{"aws": ..., "gcp": ...}` dict bypasses the `DefaultsLoaderRegistry` extension point that main provides. Any provider added after this PR would need to edit this file rather than calling `DefaultsLoaderRegistry.register("<type>", loader)` from its own registration module. After rebase, please wire GCP defaults through `DefaultsLoaderRegistry.register("gcp", GCPDefaultsLoader())` in `register_gcp_extensions` and remove the hardcoded dict.

fgogolli (commented):
Supplemental notes on items not covered by the original inline comments. The actions: rebase to pick up the `CheckHostsStatusResult` and `DefaultsLoaderRegistry` contracts, move the AWS spot-placement trio to its own AWS-only PR, add zonal MIG test coverage, fix `registration.py` to call `FieldMappingRegistry.register` and `DefaultsLoaderRegistry.register` for GCP, and drop the `PROVIDER1`/`Provider2` placeholder enum members before merge.

src/orb/providers/gcp/registration.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+        return True
+    except Exception as exc:
+        if logger:
+            logger.error("Failed to register GCP provider instance: %s", exc, exc_info=True)
+        return False
+
+
+def register_gcp_extensions(logger: Optional[LoggingPort] = None) -> None:
+    """Register GCP template extensions."""
```
fgogolli, line 161: `register_gcp_extensions` registers `TemplateExtensionRegistry` and `CLISpecRegistry` but is missing `FieldMappingRegistry.register("gcp", GCPFieldMapping())` and `DefaultsLoaderRegistry.register("gcp", GCPDefaultsLoader())`. The AWS reference implementation registers both. Without them, GCP field mappings and template defaults fall back to stubs or raise `KeyError` when the registries are iterated across all registered providers.

pyproject.toml
```diff
@@ -55,6 +55,7 @@ dependencies = [
     "requests>=2.31.0",
     "PyYAML>=6.0.0", # Exported API (lazy-loaded)
     "jsonschema>=4.17.0",
+    "google-cloud-compute>=1.39.0",
```
fgogolli: After rebase, please put the GCP SDKs under an `[gcp]` optional-dependencies extra rather than the core `[project.dependencies]` list. Main has the pattern in place: `[aws]` exists as an opt-in alias with the boto3 deps, `[all-providers]` is a meta-extra pulling them all in. Non-GCP deployments will currently pull in `google-cloud-compute` and its transitive deps (~tens of MB) on every install. Add `gcp = ["google-cloud-compute>=1.39.0"]` under `[project.optional-dependencies]` and extend `all-providers` to `["orb-py[aws,gcp]"]`.

src/orb/domain/base/value_objects.py
IkeM-L: This is not relevant to this PR? GCP does not use this, but Azure and AWS both do and this is introduced by the shared code PR not the GCP PR?

src/orb/providers/gcp/registration.py
IkeM-L: This is a pattern I copied from the AWS provider, I am happy to fix it but you should probably be aware because if it is an issue here it is an issue in AWS too?

fgogolli:
Now that #285 has merged, this PR needs a rebase with a few adjustments.

This PR carries the full #256 shared-code diff. On rebase, please drop the files that have already landed via #285 and pick up only the GCP-specific content.

The `--gcp-*` CLI flags are currently added directly to `build_parser` in `cli/args.py`. Main now routes provider-specific flags through `CLISpecRegistry`. Please add a `GCPCLISpec` class (mirror `src/orb/providers/aws/cli/aws_cli_spec.py`) that exposes the flags via `add_arguments()` and register it via the registry — the init parser injection loop picks them up automatically.

`providers/registration.py` on main uses `_REGISTERED_PROVIDERS = ["aws"]`. Please add `"gcp"` to that list instead of calling `register_gcp_provider` directly.

For `orb init --provider-type gcp` interactive setup to discover credentials, please add `get_cli_provider_config`, `test_credentials`, `get_available_credential_sources`, `get_credential_requirements` as `@classmethod` on `GCPProviderStrategy` — they are invoked before any strategy instance is constructed.

fgogolli:
This is a genuinely strong first provider contribution — thank you! The typed `GCPTemplate` subclass is exactly the right pattern, the service-account scope validator and per-operation retry profiles are really nicely done, and the MIG rollback-on-failure handling shows real care. 👏

**Sequencing heads-up (mostly automatic on rebase):**

Our prep PR (#287) lands on `main` first and renames `Template.instance_type → machine_type`. GCP's Compute Engine calls this `machineType` anyway, so renaming `GCPTemplate.instance_type → machine_type` is both the fix for the upcoming rebase *and* more idiomatic for GCP. We'll ping you when #287 is in.

1. **Package the GCP SDK as an extra** — `google-cloud-compute` is currently in core dependencies, so every non-GCP install pulls it in. Moving it to a `[gcp]` optional-extra (mirroring the existing `[aws]` pattern) keeps installs lean.
2. **Import-time side effects** — `registration.py` runs `register_gcp_extensions()` inside a module-level `with suppress(Exception)`, so simply importing the package mutates global registries (and the bootstrap also calls it, so it double-registers). Moving that into `initialize_gcp_provider()` makes it explicit and test-safe.
3. **Registry completeness** — `register_gcp_extensions()` wires `TemplateExtensionRegistry` + `CLISpecRegistry` but not `FieldMappingRegistry` / `DefaultsLoaderRegistry` (AWS registers both).
4. **CLI args** — the GCP flags are added both via `GCPCLISpec` and hand-rolled in `cli/args.py`; the `args.py` copies can go (they'll also conflict with #287's shared parser).
5. **Leftover stubs** — `PROVIDER1`/`Provider2` in the `ProviderType` enum can be dropped.

**Nice-to-haves:** a `get_resource_id_pattern()` classmethod for GCP's numeric instance IDs (new hook from #287), and dropping the extra `pycodestyle` config (repo uses ruff).

Excellent work — the structure mirrors our existing providers really well. Happy to clarify anything!

fgogolli:

- `CLISpecRegistry`
- `FieldMappingRegistry`
- `DefaultsLoaderRegistry`
- `TemplateExtensionRegistry`

PR #267, opened 2026-06-23
ci(containers): redesign image cleanup to preserve production releases

fgogolli:
## Description

Replaces the inline `container-cleanup` job inside the Container Build workflow with two dedicated workflows that distinguish production-tagged manifests from dev-tagged manifests and keep what should be kept.

**Why this is needed.** The previous cleanup step ran `actions/delete-package-versions` with `min-versions-to-keep=5` and an `ignore-versions` regex that matched any semver-shaped tag. The dev build pipeline pushes `<base>.dev<N>-python<X.Y>` for every commit on main. The Container Build's manifest creator then aliases floating tags (`python3.14`, `latest`, ...) onto the new manifest, leaving prior manifests untagged. The untagged-cleanup step then deleted those manifests, and the 5-keep limit on tagged versions only protected the very latest cycle. After a handful of dev cycles, only the most recent tag survived and older production releases were collateral-damage deleted.

**What changes.**

`container-cleanup-dev.yml` is a new workflow that runs nightly at 04:00 UTC and on `workflow_dispatch`. It classifies every container version and keeps:
- Floating tags: `latest`, `main`, `develop`, `python<X.Y>`
- Production semver tags (anything matching `v?\d+\.\d+\.\d+(...)?` that does NOT contain a `.dev<digits>` segment)
- The N most recent dev builds per Python version (default 5, configurable via input)

It deletes the rest, including untagged orphans, in the same run.

`container-cleanup-on-release.yml` is a new workflow triggered by `release: published`. It resolves the release's `published_at` timestamp and deletes dev-tagged manifests older than that timestamp, keeping a configurable grace window (default 5 per Python version) of the most recent pre-release dev builds intact for rollback. Production tags are out of scope here - they're handled by the nightly cleanup's protected list.

Both workflows support a dry-run input and write a clear plan to the job log before performing any deletions. The deletion loop tolerates per-item failures (logs and continues) so a single transient API error doesn't abort the entire pass.

The old `container-cleanup.yml` is removed because the new dev-cleanup workflow subsumes its untagged-orphan scope. The inline `container-cleanup` job in `dev-containers.yml` is also removed.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependencies update

## Related Issues
N/A

## How Has This Been Tested?

`actionlint` clean on all three changed workflows. The bash classification logic was constructed against the actual current package state (per the recent failed build's image tag `1.6.X.dev<N>-python<X.Y>` shape). End-to-end verification will run via `workflow_dispatch` with `dry-run=true` after merge; the dry-run output shows exactly what each workflow would delete before any state changes.

## Test Configuration
* Python version: N/A
* OS: GitHub Actions Ubuntu runners
* AWS region: N/A
* Dependencies changed: no

## Checklist

## Additional Notes

Both workflows use the GitHub Packages REST API directly via `gh api` rather than `actions/delete-package-versions`. That action's `ignore-versions` regex matches the whole tag string, but a single manifest can carry multiple tags (e.g. `1.6.2`, `1.6.2-python3.14`, `python3.14`, `latest`), and the keep-policy needs to consider all tags on a manifest together. The bash classification iterates over each manifest's tag list, applies the protect / dev-with-bucket / unknown rules, and decides per manifest - which is the right granularity here.

The "unknown tag" branch in `container-cleanup-dev.yml` keeps any manifest whose tags don't fit the protected or dev patterns, on the principle that an operator who manually tagged something with a custom name (e.g. `stable-2026-q2`) probably wants it kept until they say otherwise.

## Performance Impact

Both workflows run on a dedicated ubuntu-latest runner outside the build path. The API pagination + classification scales linearly with the number of container versions. Concurrent runs are serialised via a `container-cleanup` concurrency group so the dev-cleanup and on-release workflows don't race against each other.

## Security Considerations

The deletions are scoped to the package owner's container registry and require `packages: write` permission, which the default `GITHUB_TOKEN` carries. No new secrets, no new external dependencies.

## Dependencies

No new dependencies. Replaces `actions/delete-package-versions@v5` with direct `gh api` calls.

## Migration Guide

N/A.

## Deployment Notes

The first nightly run after merge will perform real deletions unless invoked manually with `dry-run=true` first. Operators wanting to verify the intended deletion set should trigger the workflow manually with `dry-run=true` before the next scheduled run.

The `container-cleanup-on-release.yml` workflow will fire automatically when the next release is published.

## Reviewers


canonicalname (approved):
Approved as discussed

PR #275, opened 2026-06-26
feat(k8s): native Kubernetes provider with Pod, Deployment, StatefulSet, and Job handlers

fgogolli:
## Description

Adds a native Kubernetes provider to ORB. The provider implements the modern `ProviderStrategy` contract and registers under the short identifier `k8s` alongside the existing AWS provider, with four handlers covering the workload-API objects operators use for compute provisioning: Pod, Deployment, StatefulSet, and Job.

The provider is gated behind an optional install extra (`pip install "orb-py[k8s]"`), so the default `pip install orb-py` install path stays lean for operators who do not run Kubernetes.

The PR also bundles a cross-cutting CLI / REST / SDK filter pass: the global `--provider` flag is split into `--provider-name` and `--provider-type`, and filter propagation is wired end-to-end across every list-style operation. Details in the "Cross-cutting CLI, REST, and SDK changes" section below.

## Why this is needed

The legacy Symphony-on-Kubernetes HostFactory plugin at `src/orb/k8s_legacy/` is in maintenance mode. Its runtime model is filesystem-as-state, with multiple cooperating watcher daemons coordinating through inotify and on-disk symlinks. This native provider supersedes it with a single in-process strategy that integrates directly with the ORB orchestration layer, the existing storage backend, the HostFactory scheduler, and the modern CQRS request lifecycle.

Beyond functional parity, this PR establishes the pattern for adding non-AWS providers: a typed Pydantic `K8sTemplate(Template)` extension, a HostFactory field-mapping registration, a CLI-spec registration, a template-extension registration, and a defaults loader. Future Azure / GCP / OCI providers can mirror this shape.

### Handlers

Four workload-API handlers, all built on a shared `K8sHandlerBase`:

- **Pod handler**: submits one bare pod per ORB machine via `CoreV1Api.create_namespaced_pod()`. Direct lifecycle control; matches the legacy plugin's primary use case and the simpler half of HPC workloads.
- **Deployment handler**: submits a `Deployment` with `spec.replicas=N` for `request.machine_count=N`. Selective machine release uses the stable Kubernetes 1.22+ `controller.kubernetes.io/pod-deletion-cost` annotation pattern: ORB annotates victim pods with a negative deletion cost, then patches `spec.replicas` down. The Deployment controller terminates the annotated pods preferentially. Suitable for long-running services and controller-managed self-heal workloads.
- **StatefulSet handler**: submits a `StatefulSet` with `spec.replicas=N`, pods named `<set>-0`, `<set>-1`, and so on. Release scales down from the highest ordinal; when the requested victims are not the highest-ordinal pods, ORB logs a warning and releases the highest ordinals to honour the StatefulSet controller's contract. Suitable for stable-hostname / persistent-storage workloads.
- **Job handler**: submits a `Job` with `spec.parallelism=N`, `spec.completions=N`, `spec.backoffLimit=0`. Release deletes the entire Job (Job parallelism is not safely mutable post-creation). Suitable for run-to-completion batch workloads.

### Configuration

`K8sProviderConfig` is a typed Pydantic model with:

- Authentication: `in_cluster: bool | None` (auto-detect via the `/var/run/secrets/kubernetes.io` sentinel), `kubeconfig_path: Path | None`, `context: str | None`
- Scope: `namespace: str = "default"`, `namespaces: list[str] | None` for multi-namespace watchers, `label_prefix: str = "orb.io"`
- Defaults: `default_node_selector`, `default_tolerations`, `default_image_pull_secret`
- Timeouts: `pod_timeout_seconds: int = 300`, `stale_cache_timeout_seconds: int = 600`
- Watch: `watch_enabled: bool = True` (daemon-mode default; CLI mode falls back to on-demand list)
- Compatibility: `emit_legacy_labels: bool = True` (also emits the legacy `symphony/open-resource-broker-reqid` label during the transition window; can be disabled once operators have fully cut over)
- Native spec: `native_spec_enabled: bool = False` (opt-in for the escape-hatch path; see below)

### Authentication

Two auth modes shipped:

- `in_cluster`: `kubernetes.config.load_incluster_config()`. Primary mode when ORB runs inside the target cluster or a management cluster with cross-cluster RBAC. Auto-detected via the in-cluster ServiceAccount-token sentinel.
- `kubeconfig`: `kubernetes.config.load_kube_config(config_file=path, context=context_name)`. Supports multiple contexts via the provider config.

OIDC and exec-plugin auth are out of scope for this initial release.

### Infrastructure discovery (`orb init --provider-type k8s`)

The provider implements the full `ProviderDiscoveryPort` interface:

- **`discover_infrastructure()`** — non-interactive discovery returning cluster topology: in-cluster flag, kubeconfig contexts, cluster endpoint, namespace list, ServiceAccounts, image pull secrets, and a three-verb RBAC probe (`create` / `watch` / `delete` pods). The return dict feeds `K8sProviderConfig` field population automatically.

- **`discover_infrastructure_interactive()`** — six-step prompt sequence: in-cluster confirmation, context selection, namespace selection (with 403 fallback to the SA-bound namespace from `/var/run/secrets/kubernetes.io/serviceaccount/namespace`), ServiceAccount selection, image pull secret selection, and RBAC probe. When any RBAC verb is missing, a pre-formatted `kubectl create rolebinding` remediation command is printed.

- **`validate_infrastructure()`** — five-check validation after `orb init`: API server reachable, context present in kubeconfig, namespace exists, ServiceAccount exists (when configured), RBAC probe passes.

`K8sProviderStrategy.get_cli_extra_config_keys()` returns `{"context", "in_cluster", "namespace", "default_image_pull_secret"}` so `init_command_handler` routes these four fields into `provider_instance.config` and `service_account` into `template_defaults`.

**RBAC remediation** messages include the full `kubectl create rolebinding` command pre-formatted with the chosen namespace and ServiceAccount, so operators can copy-paste the fix without referencing external documentation. The reference `rbac.yaml` ships both a namespace-scoped `Role` and a `ClusterRole` + `ClusterRoleBinding` variant for `namespaces: ["*"]` cluster-scoped mode.

See `docs/root/providers/k8s/discovery.md` for the full operator guide, including in-cluster and out-of-cluster deployment examples and the 403 fallback paths.

### Template extension

`K8sTemplate(Template)` is a Pydantic class extending the generic domain template with typed Kubernetes-specific fields: `namespace`, `namespaces`, `node_selector`, `tolerations`, `runtime_class`, `image_pull_secret`, `resource_requests`, `resource_limits`, `env`, `volumes`, `volume_mounts`, `pod_spec_override`, `service_account`, `native_spec`. The `service_account` field falls back to `Template.instance_profile` via a model validator so operators with cross-cloud templates do not need to set both.

K8s handlers consume the generic domain fields directly where they apply: `template.image_id` for container image, `template.tags` for Kubernetes labels (with string coercion at pod-spec time), `template.max_instances` as the per-template quota cap, `request.machine_count` for replica count. No `template.provider_data` dict reads anywhere in the provider; everything is strongly typed via `K8sTemplate`.

Template validation runs at registration time via a Pydantic-backed validator, so malformed templates are rejected before any acquire call reaches the API server.

### Native spec escape hatch

For workloads that need raw Kubernetes API fields not exposed through the typed `K8sTemplate` extension (initContainers, securityContext, affinity, hostAliases, custom controllers, etc.), the provider supports a `native_spec` field: a Jinja-templatable dict the operator sets in their template that becomes the full V1Pod / V1Deployment / V1StatefulSet / V1Job body passed to the Kubernetes API.

When the operator sets `native_spec`, it is rendered with the template + request + provider context, deep-merged over the default Jinja spec template for that handler, and submitted directly. When unset, the provider falls back to the high-level builder helpers (`build_pod_spec`, etc.) that construct the API body from the typed `K8sTemplate` fields.

The feature is opt-in via `K8sProviderConfig.native_spec_enabled` (default `False`). Default Jinja templates live under `src/orb/providers/k8s/specs/<handler>/default.json`. Setting `K8sTemplate.native_spec` with `native_spec_enabled=False` logs a `WARNING` so operators are not silently routed through the typed-builder fallback.

Server-Side Template Injection (SSTI) is mitigated by rendering through `jinja2.sandbox.SandboxedEnvironment`, which blocks dunder-traversal chains (`__class__.__mro__.__subclasses__()`), restricts unsafe global access, and prevents `undefined` escalation. The sandbox applies to both AWS and k8s providers, which share the same `JinjaSpecRenderer`. Coverage in `tests/unit/infrastructure/test_jinja_sandbox.py`.

### Watch-based status ingestion

A background asyncio watcher per provider instance streams pod events via `kubernetes.watch.Watch().stream(core_v1.list_namespaced_pod, label_selector="orb.io/managed=true", ...)` into an in-memory `K8sPodStateCache`. The cache is keyed by `(request_id, pod_name)` and is read first by `check_hosts_status`; on cold start or stale-entry fallback, a single scoped `list_namespaced_pod` call refreshes the cache.

This design avoids the per-poll list call on busy clusters that would otherwise pressure the API server. The watch loop handles 410 Gone by reconnecting with `resource_version=None` and applies exponential backoff on disconnect, matching the production-hardened pattern from the legacy plugin. The circuit-breaker in the shared infrastructure resilience layer prevents cascading failures when the API server is degraded.

For CLI mode (no long-running process), the watcher is not started; `check_hosts_status` falls back to direct list calls. `cleanup()` blocks on watcher shutdown with a timeout, so sockets are closed before the process continues past cleanup.

### Startup reconciliation and orphan garbage collection

On daemon start, a reconciler lists all pods cluster-wide (or namespace-scoped per config) with `orb.io/managed=true`, cross-references against ORB request records, and rebuilds the in-memory cache. Pods without a matching request record are tagged as orphans.

A periodic orphan-GC task (default every 300 seconds) lists orphan pods and, if `auto_cleanup_orphans=True`, deletes them. The GC checks `pod.creation_timestamp` against `orphan_gc_interval_seconds + buffer` before any deletion, preventing in-flight pods (created between the acquire write and the storage commit) from being immediately deleted. Conservative by default: orphans are logged and metric-reported but not deleted unless the operator explicitly enables cleanup.

A per-request timeout-GC integrated with `check_hosts_status` marks pods stuck in `Pending` past `pod_timeout_seconds` as `TERMINATED` with the unschedulable reason surfaced in `provider_data.unschedulable_reason`, and issues `delete_namespaced_pod` once the threshold is exceeded so pods do not accumulate.

### Observability

- **Node state tracking**: `K8sPodStateCache` records the node name, availability zone (via `topology.kubernetes.io/zone`), and instance type (via `node.kubernetes.io/instance-type`) alongside each pod entry, surfacing per-machine placement metadata in `check_hosts_status` responses without additional API calls.
- **Disruption and restart signals**: `extract_status_reason` detects `DisruptionTarget` pod conditions (Karpenter drains, preemptions) and `PodState` tracks `restart_count`, closing observability gaps versus the legacy plugin.
- **Pod-spec audit on submission**: The spec builders inspect the incoming `K8sTemplate` (and `native_spec` when enabled) for high-risk fields — `hostPath` volumes, `privileged` containers, `hostNetwork`, and dangerous capabilities (`NET_ADMIN`, `SYS_ADMIN`, `SYS_PTRACE`) — and emit a structured `WARNING` log entry per violation. No requests are blocked; the audit gives operators visibility without imposing policy.
- **Health**: `HealthCheckPort` is registered via `register_k8s_health_checks` mirroring the AWS pattern, so k8s health checks appear on the `/health` endpoint.
- **Templates surface**: `K8sTemplateExampleGeneratorAdapter` is registered so `orb templates` surfaces k8s examples through the DI path.

### HostFactory integration

`K8sFieldMapping` registers with the existing `FieldMappingRegistry`, translating Symphony HostFactory camelCase JSON fields to internal snake_case names: `containerImage`, `namespace`, `resourceRequests`, `resourceLimits`, `runtimeClass`, `nodeSelector`, `tolerations`, `serviceAccount`, `imagePullSecret`, `completions`, `parallelism`, `replicas`, `labels`, `annotations`, `volumeMounts`, `volumes`, `env`. HF JSON sets a `namespace` value that takes precedence over the provider config default.

A self-contained `hostfactory/providers/k8s-hf/scripts/` directory ships with the provider, giving HostFactory-integrated deployments the same `requestMachines`, `getRequestStatus`, `getReturnRequests`, `getAvailableTemplates`, and `requestReturnMachines` shell entry points already present for the AWS provider.

### Phase mapping

Pod phase `Succeeded` resolves to `COMPLETED` for Job handler pods and to `TERMINATED` for Pod, Deployment, and StatefulSet handler pods, matching the expected lifecycle semantics for each workload type.

### Retry semantics

`ExponentialBackoffStrategy.should_retry` inspects the HTTP status code and skips retry on 409 (conflict), 400, 403, and 422, so RBAC and idempotency errors surface immediately instead of running through the retry budget.

### Plugin discovery for CRD-extending providers

`src/orb/providers/registration.py` adds a `discover_provider_plugins()` function that consults `importlib.metadata.entry_points(group="orb.providers")` and registers any third-party packages that advertise themselves there. This is the documented extension path for CRD-specific providers (Kubeflow MPI Operator, PyTorch Operator, Spark Operator, Volcano, etc.): each CRD provider ships as its own PyPI package, registers via the entry point, and extends `K8sProviderStrategy` rather than duplicating its base.

Built-in providers (`aws`, `k8s`) remain hardcoded in `_REGISTERED_PROVIDERS`; entry-point auto-discovery is for external packages.

## Cross-cutting CLI, REST, and SDK changes

These changes are provider-agnostic and apply equally to AWS and future providers.

### Flag surface

The single global `--provider` flag is replaced by two orthogonal flags. `--provider` accepted only a provider **instance name**, which made "operate on all k8s instances" impossible without listing every name. The split:

- `--provider-name NAME` restricts the operation to one specific provider instance by exact name (e.g. `aws-prod-us-east-1`).
- `--provider-type TYPE` restricts the operation to every active instance of a given type (`aws`, `k8s`, etc.).

Both are inherited via `add_global_arguments` so every subcommand that reads provider-scoped state accepts them. `orb init --provider TYPE` is renamed to `orb init --provider-type TYPE` for consistency. The rename is a clean break with no shim; existing scripts must migrate.

The generic `--filter` flag (previously declared but not consumed by list handlers) now populates `filter_expressions` on every list operation and reaches the query handler.

### End-to-end wiring

Filter chain now flows CLI / REST / SDK → Input DTO → Query DTO → handler for every list-style operation:

- `list_machines`, `stop_machines`, `start_machines`, `return_machines`, `get_machine_status` (all-mode)
- `list_requests` (sync and async paths), `list_return_requests`
- `list_templates`
- `list_providers`, `provider_health`

Every Input DTO gains `provider_name`, `provider_type`, and `filter_expressions` fields where they apply. Every orchestrator forwards them into the corresponding Query DTO. Every handler applies them. `ProviderRegistry.select_active_provider` reads both overrides from `ConfigurationPort` with name-first precedence.

### REST

`GET /api/v1/machines/`, `/api/v1/requests/`, `/api/v1/requests/returns`, `/api/v1/templates/` all gain `provider_name`, `provider_type`, and `filter_expressions` query params (repeated `?filter_expressions=a&filter_expressions=b`). `GET /api/v1/machines/` also gains `timestamp_format`. `GET /api/v1/requests/` also gains `template_id` and `request_type`. `POST /api/v1/machines/return` request body gains `provider_name` and `provider_type` alongside `machine_ids` / `all_machines` / `force`.

### SDK

`SDKConfig` gains `provider_name` and `provider_type` fields with `ORB_PROVIDER_NAME` and `ORB_PROVIDER_TYPE` env fallbacks, applied to `ConfigurationPort.override_provider_name` / `override_provider_type` on `ORBClient.initialize()`. Per-call kwargs on `list_machines`, `list_requests`, `list_return_requests`, `list_templates`, `return_machines`, `stop_machines`, and `start_machines` accept both filters; when a kwarg is omitted, the SDKConfig default applies. The legacy `SDKConfig.provider` string is left untouched for backward compatibility.

### Bug fixes

- `ListRequestsHandler.execute_query` compared the `provider_name` filter against `r.provider_api` (wrong attribute). Fixed to compare against `r.provider_name` for exact-match semantics matching every other handler.
- `ListMachinesOrchestrator` dropped `timestamp_format` when constructing `ListMachinesQuery`. Fixed.
- `ListRequestsOrchestrator` async path dropped `request_type`. Fixed.
- `ListActiveRequestsQuery` was missing `template_id`, so the sync path bypassed the filter that the async path applied via in-memory post-filter. Both paths now behave identically.
- Dead `--generic` flag on `templates generate` is removed. `--summary` on `infra discover` is retained (backed by `_discover_infrastructure_summary`).

## Bootstrap and quality fixes

- `Application(config_path=X)` ignored `X` when `config_dict` was `None`, silently falling back to the DI-registered default `ConfigurationManager()` with its own file-discovery. Explicit `config_path` now always wins.
- `StartupValidator._find_config_file` used a bespoke candidate list containing only `./config/config.json` as its fallback, missing `~/.orb/config/config.json` and every other platform root that `ConfigurationManager` resolves via `platform_dirs.get_config_location()`. The validator now routes through `get_config_location()` and adds a final user-home fallback for operators running from inside a repo checkout, where `platform_dirs` pins the config location to the venv root regardless of whether config exists there. Full consolidation of the divergent path-discovery implementations is tracked as a follow-up.
- Six `except` blocks with implicit `pass` bodies flagged by static analysis are rewritten with explicit fall-through returns or `contextlib.suppress`, matching the intent already documented in their surrounding comments.

## Type of Change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependencies update

## Related Issues
N/A

## How Has This Been Tested?

End-to-end shim tests at `tests/integration/providers/k8s/` invoke the modern `orb` CLI against fully mocked Kubernetes client APIs and exercise the full acquire → status → release flow per handler, plus the watch lifecycle (start, 410-Gone reconnection, cache population), the startup reconciler, and the HostFactory field-mapping round-trip. Unit tests cover each handler, the cache, the watcher, the field mapping, the template extension, the auth resolvers, and the orphan GC. The full modern test suite passes unchanged.

Local verification:
- `pytest tests/unit tests/integration -q --no-cov`: green
- `pyright src/`: 0 errors
- `ruff format --check --quiet .`: clean
- `ruff check --select W,F,I --ignore E501 --quiet .`: clean
- `shellcheck` over all shell scripts: clean
- `orb providers list`: surfaces `k8s` alongside `aws`
- `orb k8s-legacy --help`: legacy shim continues to work unchanged

## Test Configuration
* Python version: 3.12
* OS: GitHub Actions Ubuntu 24.04, local macOS Darwin 25.3.0
* AWS region: N/A
* Dependencies changed: yes. A new `[k8s]` optional install extra adds `kubernetes`, `fastapi`, `uvicorn`, `sqlalchemy`, `alembic`, `pg8000`, `prometheus-client`, `watchdog`, `tenacity`, `rich`, `jinja2`, `pydantic-settings` to the dependency set when explicitly requested. The bare `pip install orb-py` path is unaffected.

## Checklist
- [x] I have made corresponding changes to the documentation

## Additional Notes

Open design points documented in the migration guide for operator consideration:

- The watcher runs in-process by default. For pure CLI-mode deployments (no long-running ORB daemon), the watcher is not started and `check_hosts_status` falls back to scoped list calls per poll.
- RBAC is the operator's responsibility. A reference manifest is provided at `docs/root/providers/k8s/rbac.yaml`. ORB does not programmatically create ServiceAccount / Role / RoleBinding resources at startup.
- The provider supports a single namespace per provider instance by default. Multi-namespace deployments register `namespaces: ["a", "b", ...]` or `["*"]` for cluster-scoped watchers.
- StatefulSet release falls back to highest-ordinal termination when requested victims are not the highest ordinals. Operators relying on strict per-pod release should use Deployment or Pod handlers.
- Job release deletes the entire Job; per-pod release within a Job is not supported (Job parallelism is not safely mutable post-creation).

## Performance Impact

Watch-based status ingestion is the load-bearing design choice: it avoids the per-poll list-pods API server pressure that would otherwise cap throughput on busy clusters. The in-process watcher uses one thread per provider instance per resource type and reconnects via exponential backoff on 410 Gone or transient disconnect.

## Security Considerations
- [x] Security implications reviewed and addressed

The `native_spec` field accepts operator-supplied Jinja template strings that are rendered server-side. To mitigate Server-Side Template Injection (SSTI), the Jinja-based spec renderer runs inside `SandboxedEnvironment` to block native-spec template injection. The sandbox disallows dunder attribute traversal (blocking the `__class__.__mro__.__subclasses__()` RCE chain), restricts unsafe global access, and prevents `undefined` escalation. This affects both the AWS and k8s providers, which share the same `JinjaSpecRenderer`. Coverage is in `tests/unit/infrastructure/test_jinja_sandbox.py`. The `native_spec_enabled` flag defaults to `False`; operators explicitly opt in.

The provider operates within the credentials supplied by the operator (in-cluster ServiceAccount or kubeconfig). No new secrets, no new external network actions beyond the Kubernetes API server.

## Dependencies

Adds `[project.optional-dependencies] k8s = [...]` with the heavy Kubernetes deps. No mandatory changes to the bare-install path.

## Migration Guide

See `docs/root/providers/k8s/migrating-from-k8s-legacy.md` for the complete migration path from the legacy plugin (`orb k8s-legacy ...`) to the modern provider (`orb requests submit --provider-name k8s ...`). Field mapping table, label conventions, behaviour deltas, and coexistence strategy covered.

## Deployment Notes

The first time an operator routes traffic to `--provider-type k8s`, they should verify:
1. RBAC is in place (apply the reference manifest at `docs/root/providers/k8s/rbac.yaml`)
2. `kubeconfig` is reachable, or ORB is running in-cluster with a ServiceAccount that has the required permissions
3. `K8sProviderConfig.namespace` (or `namespaces`) points at the intended target
4. If using the `native_spec` escape hatch, `native_spec_enabled: True` is set in the provider config

## Reviewers


canonicalname (dismissed):
Approved

canonicalname (approved):
Approved as discussed

PR #286, opened 2026-07-07
feat(k8s): provider hardening — perf, correctness, observability, live-test coverage

fgogolli:
## Description

Post-merge hardening of the native Kubernetes provider.  Fixes watch and reconciliation correctness gaps, tightens hot-path performance, wires a Prometheus metric surface (matching the legacy provider's metric names) into the handler and watcher hot paths, and adds fourteen live-cluster test scenarios.  Every production change carries at least one unit test; live tests are gated behind a `k8s_live` marker and `--run-k8s` flag so nothing runs against a cluster by default.

Highlights:

- **Watch correctness** — full LIST after 410-Gone (rv=0 was silently dropping events between the stale rv and the cache start).  `is_healthy()` returns False until the first successful sync completes; `MultiNamespaceWatcher.stop()` tolerates per-namespace failures.
- **Reconciliation correctness** — startup reconciler failures no longer mark first-sync-complete, so `is_healthy()` cannot report True with a cold empty cache.  Orphan GC and StartupReconciler fan out namespaces in parallel via `asyncio.gather`.
- **Performance** — pod-state-cache `mark_stale` uses per-key locks; StatefulSet ordinal parse replaces regex with rsplit.
- **Circuit breaker** — base class exposes a `_get_failure_threshold()` hook.  K8sCircuitBreaker overrides only that method to consult a live callable so config-reloads take effect without a process restart, with safe fallbacks on raise / non-int return.
- **Metrics (wired, not scaffolded)** — nine Prometheus metrics matching the legacy k8s exporter, registered on the default `REGISTRY` so the standard `/metrics` scrape sees them.  Handlers record acquire/release counts and pod-creation outcomes on every `acquire_hosts`/`release_hosts`; the watcher records watch events per apiserver verb and reconnects per reason.  Enum-guarded label helpers keep cardinality bounded.  Gated by `K8sProviderConfig.metrics_enabled` (default True).
- **Handlers** — `delete_one_pod` collapses double-try into a single retry-aware path; deployment annotation-abort uses `math.ceil` (exact 50% aborts) and no-ops on empty pod_names.
- **Contract test** — kmock `deleted_pods` closure state no longer leaks between two acquire calls in the shared-adapter stability scenario.
- **Live scenarios** — 14 new files covering watch 410-Gone, REST + Go SDK cycles, token expiry, RBAC revocation, ResourceQuota rejection, partial fulfilment, cross-request isolation, concurrent acquire/release, karpenter cold-node, terminal-state restart, metric emission, ARM64 nodeAffinity, watch buffer overflow.

## Type of Change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Dependencies update

## Related Issues
None.

## How Has This Been Tested?

`pyright src/orb/providers/k8s src/orb/infrastructure/resilience` — 0 errors.  `pytest tests/providers/k8s/unit tests/providers/k8s/contract tests/integration/providers/k8s` — 1124 pass; 3 pre-existing failures unrelated to this branch stay deselected.  `pytest tests/providers/k8s/live --collect-only` — collects cleanly.  `ruff check --select W,F,I` and `ruff format --check` — clean across the repo.

New parallelism proofs (barrier-based) fail on serial regressions — the earlier "all namespaces called" assertions would have silently passed a for-loop wrapped in `asyncio.gather`.

Live-cluster scenarios have been authored and collect cleanly but have not yet been exercised against a real cluster in CI.

## Test Configuration
* Python version: 3.12
* OS: Darwin 25 (macOS) locally; Linux in CI
* AWS region: n/a
* Dependencies changed: none

## Checklist

## Additional Notes

The Prometheus metric surface is fully wired into the handler and watcher hot paths in this PR — it is not a scaffold.  A follow-up epic tracks consolidating this metric stack with the AWS provider's metric collector behind a shared, OpenTelemetry-compatible port; that consolidation is out of scope here.

Live scenarios that depend on cluster capabilities beyond a vanilla kubeadm cluster (karpenter, ARM64 nodes, Prometheus scrape endpoint, RBAC admin) skip cleanly with a diagnostic message when the prerequisite is missing.  `test_watch_survives_burst_event_flood` cleans up its burst pods in `try/finally` — a failing post-burst assertion no longer orphans workload on the cluster.

n/a

## Performance Impact
- [x] Performance improved
- [ ] No significant performance impact

Orphan GC and startup reconciler fan-out is now wall-clock bounded by the slowest namespace instead of the sum of all namespaces.  Pod-state-cache `mark_stale` no longer serialises distinct keys on a single global lock.  StatefulSet ordinal parse skips regex compilation on the status-check hot path.

## Security Considerations
- [x] Security improved

Prometheus label values for `reason`, `status`, and `event_type` are pinned to frozensets of allowed enum strings via `record_watch_reconnect` / `record_pod_creation` / `record_watch_event` helpers.  Unknown values bucket to `"unknown"` rather than creating new time series — closes an unbounded-cardinality vector.  The circuit-breaker threshold callable is wrapped in try/except so a raising or non-int provider no longer disables the breaker (previously would let unbounded failures accumulate before opening).

## Dependencies
None.

## Migration Guide
Not a breaking change.  Consumers that construct `K8sMetrics()` without passing a registry now register against the process-wide `REGISTRY` — this is the correct production behaviour but callers that previously relied on registry isolation must pass an explicit `CollectorRegistry`.  All tests in this repo have been updated to pass one; external callers should audit.

## Deployment Notes
The k8s provider now emits Prometheus metrics on the standard `/metrics` scrape endpoint when `metrics_enabled` is True (the default).  Operators scraping the endpoint will see new `orb_k8s_*` series.  Set `metrics_enabled: false` in the provider config to opt out.

## Reviewers

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -492,6 +492,7 @@ markers = [
     "api: marks API tests",
     "security: marks security tests",
     "serial: forces serial execution (use ``-n0`` or filter with ``-m 'not serial'`` under xdist)",
+    "k8s_live: Live Kubernetes integration tests (require a real cluster — pass --run-k8s to enable)",
 ]
 # Provider-specific fake credentials live in
 # tests/providers/<name>/mocked/conftest.py.  The global pytest-env block
--- a/src/orb/api/server.py
+++ b/src/orb/api/server.py
@@ -507,8 +507,10 @@ def create_fastapi_app(server_config: Any) -> Any:
             from prometheus_client import REGISTRY, generate_latest
 
             registry_text = generate_latest(REGISTRY).decode("utf-8")
-        except Exception:  # ImportError or any prometheus_client internal error
-            pass
+        except Exception:  # noqa: BLE001 — ImportError or prometheus_client internal error
+            # prometheus_client is an optional [monitoring] extra; a minimal
+            # install without it must still serve the homegrown collector text.
+            registry_text = ""
 
         body = homegrown_text
         if registry_text:
--- a/src/orb/api/server.py
+++ b/src/orb/api/server.py
@@ -495,12 +495,26 @@ def create_fastapi_app(server_config: Any) -> Any:
 
         try:
             collector = get_container().get_optional(MetricsCollector)
-            if collector is None:
-                return Response(content="", media_type="text/plain; version=0.0.4")  # type: ignore[misc]
-            prometheus_text = collector.to_prometheus_text()
-            return Response(content=prometheus_text, media_type="text/plain; version=0.0.4")  # type: ignore[misc]
+            homegrown_text: str = collector.to_prometheus_text() if collector is not None else ""
         except Exception:
-            return Response(content="", media_type="text/plain; version=0.0.4")  # type: ignore[misc]
+            homegrown_text = ""
+
+        # Append prometheus_client REGISTRY output (k8s metrics live there).
+        # prometheus_client is an optional [monitoring] extra — guard the import
+        # so a minimal install without it still returns the homegrown text.
+        registry_text: str = ""
+        try:
+            from prometheus_client import REGISTRY, generate_latest
+
+            registry_text = generate_latest(REGISTRY).decode("utf-8")
+        except Exception:  # ImportError or any prometheus_client internal error
+            pass
+
+        body = homegrown_text
+        if registry_text:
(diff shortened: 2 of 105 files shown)


canonicalname (approved):
Approved as discussed over fruit.

PR #320, opened 2026-07-18
test: raise combined coverage to ~86% (+ fix 3 latent bugs surfaced by the new tests)

fgogolli:
## Description

Raises combined test coverage from 66.88% to ~86% line-rate (84.46% branch-inclusive; part of the lift is genuine new tests, part is omitting Reflex render trees from the denominator) (unit + integration + provider legs), adding and strengthening unit tests only. Along the way the new tests surfaced three latent source bugs, which are fixed here, and a few CI/coverage-config issues, which are corrected.

The branch is organised into three logical commits:
1. **`fix(orb)`** — three latent source bugs the new coverage exposed.
2. **`ci(coverage)`** — the combined-coverage gate, UI-render omit, and Codecov alignment.
3. **`test`** — the coverage work itself.

**Source bugs fixed** (each had a test that previously asserted *around* the broken behavior; those tests now assert the corrected behavior):
- `template_query_handlers` — the template load-error path raised `UnboundLocalError` because `EntityNotFoundError` was imported inside a conditional block (making it function-local); hoisted to module scope.
- `ui/api_http init_orb` — a caller-supplied `confirm` key could override the mandatory `INIT` safety token (`{"confirm":"INIT", **body}`); the token is now pinned last so it stays authoritative.
- `storage_command_handlers` — `handle_validate_storage_config` imported `GetStorageHealthQuery` from the wrong module, so the import always failed and the handler always reported "valid" via a dead fallback; fixed the import so validation actually runs.

**Test quality** — an adversarial review of the added tests flagged weak/tautological assertions and change-detectors; those were rewritten to assert real behavior (return values, call arguments, specific exception types), not shapes. Tests are isolated for parallel (xdist) runs: global singletons (DI container factory, domain container, provider registration, AWS template extension) are reset between tests, async helpers use fresh event loops, and optional-dependency tests skip cleanly when `rich`/`psutil` are absent (as they are in the minimal CI unit env).

**Coverage config** — Reflex UI render trees (`ui/pages`, `app`, `rxconfig`) are omitted from coverage in both `pyproject.toml` and `codecov.yml`, mirroring the existing `k8s_legacy` exclusion: they are exercised by the `ui-smoke` leg and the Reflex compile step, not by pytest. Codecov is configured to read only the single omit-applied combined report (`disable_search`) so its number tracks the gate. The gate is set to an 80% floor.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement

## Related Issues
N/A — test-coverage initiative.

## How Has This Been Tested?
- [ ] Manual testing performed

Both CI legs run green under xdist: unit ~12,795 passed, providers ~4,354 passed. The three source-bug fixes are covered by tests that now assert the corrected behavior. Optional-dependency behavior verified with `rich`/`psutil` hidden (they skip cleanly). Combined coverage: 86.56% line-rate (43962/50790), 84.46% branch-inclusive; gate enforces the branch-inclusive metric at an 80% floor. Baseline on main was 66.88%.

## Test Configuration
* Python version: 3.12 / 3.14 (CI)
* OS: macOS (darwin), Linux CI
* AWS region: n/a (mocked)
* Dependencies changed: none

## Checklist

## Additional Notes

The coverage gate uses coverage.py's branch-inclusive total (~84%); Codecov reports the XML line-rate (~86%). The gate now prints both from the same combined dataset so the two numbers can be reconciled from the run log. The `omit`/`ignore` lists reflect real testable-code coverage — render trees and the legacy k8s plugin (own suites) are excluded, not counted at 0%.

n/a

## Performance Impact
- [ ] Performance improved

## Dependencies
None.

## Migration Guide
The only behavioral change for consumers is CI: a combined-coverage gate (80%) is now enforced; a future PR that drops coverage below it will fail the check.

## Deployment Notes
None.

## Reviewers

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -360,6 +360,11 @@ jobs:
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         files: coverage-combined.xml
+        # Only upload the single omit-applied combined report.  Without this,
+        # the uploader also auto-discovers the per-leg XMLs under
+        # coverage-artifacts/ and merges them (raw, pre-omit, overlapping),
+        # producing a different — lower — number than the coverage gate.
+        disable_search: true
         name: combined
         # Overwrite any prior partial report for this commit with the full merge.
         override_commit: ${{ github.event.pull_request.head.sha || github.sha }}
--- a/.project.yml
+++ b/.project.yml
@@ -23,7 +23,7 @@ build:
   package_root: src/orb
 ci:
   test_timeout: 300 # Test timeout in seconds
-  coverage_threshold: 60 # Minimum combined coverage percentage (ratchet up as coverage improves)
+  coverage_threshold: 80 # Minimum combined coverage percentage (ratchet up as coverage improves)
 development:
   pre_commit_enabled: true
   security_scan_enabled: true
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,7 +7,7 @@ coverage:
   status:
     project:
       default:
-        target: 60% # Combined-coverage floor across all test legs (ratchet up as coverage improves)
+        target: 80% # Combined-coverage floor across all test legs (ratchet up as coverage improves)
         threshold: 0%
         informational: false
     patch:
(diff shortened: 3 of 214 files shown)


canonicalname (approved):
LGTM

PR #328, opened 2026-07-20
refactor(cli): consolidate argument schema with intent-based parent parsers

fgogolli:
## Description
Consolidates the CLI argument schema around a small set of intent-based **ParentParsers** (common, list, write, provider-scope, hf-compat). Subcommands are now composed from these shared parsers instead of ~40 hand-written `add_global_arguments` calls.

Because flags are declared once on a shared parent and composed in, duplicate flags now raise at parser **build-time** rather than surfacing as confusing conflicts at parse-time. This PR also fixes the long-standing `-f`/`--file` destination collision, consolidates the provider filter into a single request-scoped conduit, tightens `--all` scoping, and single-sources config-file discovery.

### What / Why
- **Intent-based parent parsers** — replace ~40 manual `add_global_arguments` calls; each flag is declared once, so collisions fail fast at build-time.
- **`-f`/`--file` dest collision** — each parser now uses a single, unambiguous `dest`, removing the ambiguity between the two flags.
- **Provider-filter conduit** — the provider filter now travels on the query/command DTO (request-scoped) instead of being smuggled through container state. This removes the `ConfigurationPort` override methods and the matching `ConfigurationManager` private fields.
- **`--all` scoping** — moved from the global flag set to per-command opt-in for commands that actually support it.
- **Config-file discovery** — single-sourced through `platform_dirs.resolve_config_file()`.

## Type of Change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Related Issues
N/A

## How Has This Been Tested?

### Intentional CLI behavior changes
- `--all` is **no longer a global flag**; it is now per-command opt-in for commands that support it.
- `--yes` is now **scoped to write commands** only.
- The provider filter is passed on the **request DTO**, no longer via shared container state.

## Checklist

## Migration Guide
- Scripts relying on `--all` as a global flag must move it to the specific command that supports it.
- `--yes` is only accepted on write commands now.
- Any tooling that relied on the provider filter being set via container/configuration state should pass it as part of the request instead.

## Performance Impact

## Security Considerations

## Reviewers

### DIFF
--- a/src/orb/cli/args.py
+++ b/src/orb/cli/args.py
@@ -42,56 +42,114 @@ except ImportError:
     HELP_FORMATTER = argparse.RawDescriptionHelpFormatter
 
 
-def add_global_arguments(parser):
-    """Add arguments that should be available on all commands."""
-    parser.add_argument(
-        "-f",
-        dest="hf_file",
-        metavar="FILE",
-        help="Input JSON file path (HostFactory compatibility)",
-    )
-    parser.add_argument(
-        "-d",
-        dest="hf_data",
-        metavar="DATA",
-        help="Input JSON data string (HostFactory compatibility)",
-    )
-    parser.add_argument(
-        "--provider-name",
-        dest="provider_name",
-        metavar="NAME",
-        help="Restrict to a specific provider instance by exact name (e.g. aws-prod)",
-    )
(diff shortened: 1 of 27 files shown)


canonicalname (approved):
Nice

PR #332, opened 2026-07-20
feat(request): fulfilment state machine, plus provider/interface/k8s-auth hardening

fgogolli:
## Description
Introduces a single state-machine write-authority for request status, then hardens the surrounding fulfilment, provider, and interface paths uncovered while wiring it in. Every status transition now flows through one place, giving consistent invariants and preserving domain events across mutators.

Fulfilment state machine:
- **PARTIAL_PENDING holding state with a lazy deadline (`deadline_at`)** — a partially-fulfilled acquire gets a grace window before resolving to a terminal `PARTIAL` or `TIMEOUT`, instead of settling immediately.
- **`FulfilmentDiagnostic` value object** — captures *why* a request ended partial, failed, or cancelled, and surfaces that reason on the request and on wire responses.
- **Provider error classification** — maps provider errors to diagnostic categories that feed the diagnostic attached to the request, via a pure-domain classifier the AWS provider delegates to.
- **ACQUIRE vs RETURN semantics separated** — a RETURN keeps its terminal partial outcome rather than being re-evaluated like an acquire.
- **Explicit partial finality** — a settled partial is signalled with an explicit `final` flag rather than inferred from `pending_count`, and that verdict is propagated through multi-fleet aggregation (correct `target_units`, diagnostics gated to shortfall states).
- **Startup recovery** — sweeps expired active requests through the state machine so deadlines are honoured across restarts; the sweep runs after the server starts serving so it does not block readiness.

Correctness fixes surfaced alongside:
- **Kubernetes auth is now terminal-independent** — the k8s client's exec-credential plugin previously ran interactively whenever stdout was a TTY, which for `credential_process`-backed EKS profiles failed to attach the bearer token and returned 401 from an interactive shell (but not when output was redirected). The exec token mint — including the SDK's lazy per-request refresh — now always runs non-interactively.
- **Template default merge is alias-aware** — merging default layers no longer lets a canonical field and its deprecated alias both survive into the model, which could pick the wrong value; all default layers (including provider-subclass fields and extension layers) coalesce aliased fields into one slot.
- **Init credential inquiry** resolves the provider strategy correctly, so credential-source / requirement inquiries work instead of reporting the provider unsupported.
- **Pagination fields** (`next_cursor`, `total_count`) are emitted only on list responses, not on single-status or Host Factory payloads.
- **Optimistic-concurrency retry limit** is configurable instead of hard-coded.

## Type of Change

## Related Issues
N/A

## How Has This Been Tested?

Unit, integration, provider, architecture, and CQRS suites pass; pyright and the CI ruff profile are clean. The k8s provider's live-cluster suite (behind `--run-k8s`) passes against a real EKS cluster, including the previously terminal-only auth path.

## Test Configuration
* Python version: 3.12 (CI matrix 3.10–3.14)
* OS: macOS / Linux CI
* AWS region: n/a (no live AWS in CI)
* Dependencies changed: none

## Backward Compatibility
The Alembic migration adds four **nullable** columns to `requests` (`deadline_at`, `partial_since`, `last_transition_at`, `fulfilment_diagnostic`) with no server default. Existing rows read back as `NULL`, so the change is purely additive and the migration is fully reversible. The template field renames remain backward compatible via alias choices and deprecated read-only properties.

## New Request States
`PARTIAL_PENDING` is added as an intermediate holding state between active provisioning and the terminal `PARTIAL`/`TIMEOUT` outcomes. Terminal states are unchanged in meaning; RETURN flows retain their terminal partial result.

## Checklist
- [x] I have added tests that prove my fix is effective and my feature works

## Additional Notes
The k8s auth fix keeps the process-global stdout override scoped to the credential mint alone; it is never held across a blocking watch stream. In-cluster (service-account) and static bearer/cert auth are unaffected.

## Screenshots
N/A

## Performance Impact

## Security Considerations

The k8s auth change makes credential minting non-interactive; it does not weaken any authentication or authorization check.

## Dependencies
None.

## Migration Guide
Apply the Alembic migration to add the four nullable `requests` columns; no data backfill or downtime is required, and it is reversible. No caller changes are required.

## Deployment Notes
No configuration changes are required to deploy. Operators running the k8s provider from an interactive terminal will no longer see spurious 401s from the credential plugin.

## Reviewers


canonicalname (dismissed):
ok

kirillsc (approved):
re-approved

PR #336, opened 2026-07-24
refactor: AWS auth config DRY, deprecation signals, resilience dependency inversion, CLI provider-type guard

fgogolli:
## Description
Finishes several lingering quality/architecture cleanups.

- **AWS auth config DRY** — the IAM, Cognito, and session-factory clients each built their own botocore `Config` with identical timeouts and retry count. They now route through the shared `get_boto3_config()` helper (one source of truth), which also brings their retry mode in line with every other client in the codebase (adaptive; the inline configs had left it at the implicit legacy default).
- **Deprecation signals** — several operator-facing config surfaces accepted legacy keys silently (k8s + k8s-DTO `env` alias, AWS connect-timeout key remap, k8s Symphony-era key remap). Each now emits a `logger.warning`, matching the house standard: operator surfaces warn via the logger, developer/SDK surfaces via `DeprecationWarning`.
- **Resilience dependency inversion** — the provisioning orchestration service imported a concrete circuit-breaker error from infrastructure, an inward layer-boundary violation held open by an import-linter exception. A resilience port (a `CircuitBreakerPort` protocol + `CircuitBreakerOpenError` base) now lives in the application layer; the service depends on that, and the import-linter exception is removed.
- **CLI provider-type guard** — a regression test asserting `--provider-type` is declared exactly once per leaf sub-parser (inherited from the shared parent parser) and that registering an additional provider spec merely widens its `choices` rather than adding a second conflicting option string. Keeps the CLI stable as new provider types come online.

## Type of Change

## How Has This Been Tested?
- pyright: 0 errors on all changed files. import-linter: 3 contracts kept, 0 broken — "Application layer does not import infrastructure concretes" passes with the exception removed. Targeted pytest subsets (boto/config/iam/cognito/session; deprecation/alias/config/sdk; CLI provider-type) pass.

## Test Configuration
* OS: ubuntu-latest
* Dependencies changed: none

## Checklist

## Security Considerations

## Additional Notes
The AWS auth clients' retry mode moves from the implicit legacy default to adaptive as a side effect of the config consolidation — a strictly better throttling policy for STS/IAM/Cognito, consistent with the rest of the codebase; no timeout or attempt-count values change.

## Reviewers

### DIFF
--- a/.importlinter
+++ b/.importlinter
@@ -46,13 +46,11 @@ allow_indirect_imports = True
 #   A2 — same fix as A1
 #   A3 — MetricsCollector now injected via MetricsPort
 #
-# One pre-existing violation remains and is deliberately ignored below: the
-# provisioning orchestration service still reaches for the circuit-breaker
-# resilience concretes.  Inverting this requires a domain-level
-# CircuitBreakerError base + a resilience port, tracked by the outstanding
-# application→infrastructure ports work.  It is ignored (not silently allowed)
-# so this contract stays honest — a green result here means "no NEW app→infra
-# leaks", and any additional violation will still break the build.
+# The circuit-breaker resilience leak has also been inverted: the orchestration
+# service now depends on orb.application.ports.resilience_port (a
+# CircuitBreakerPort protocol + CircuitBreakerOpenError base) and the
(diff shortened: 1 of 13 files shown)


canonicalname (approved):
LGTM

PR #338, opened 2026-07-24
chore: quality sweep — 404 not-found fix, security hardening, boto DRY, dead-code + test + docs cleanup

fgogolli:
## Description
A batch of correctness, security, and quality cleanups across the API, AWS provider, SDK, storage, MCP, and CI surfaces.

### Bug fixes
- **404 for missing requests** — `GET /requests/{id}` and `/requests/{id}/status` returned HTTP 200 with an empty-but-valid body for a non-existent request ID. The shared status orchestrator collapsed every per-ID exception into one `{request_id, error}` shape, so the endpoint could not tell "not found" from "exists but provider-sync errored". The orchestrator now tags a genuine `EntityNotFoundError` with an explicit `not_found` flag (other per-ID errors keep a plain error entry so the batch endpoint still reports partial failures), and the single-ID endpoints raise `EntityNotFoundError` → 404 only on that flag. A request that exists but failed provisioning (status set, populated error block) still returns 200 with its detail.

### Security
- **Host-header protection** — the server now logs a startup WARNING when `trusted_hosts` is empty or contains `*` (either of which disables `TrustedHostMiddleware`), so an operator is told when DNS-rebinding/Host-spoofing protection is off.

### Correctness / provider-agnostic
- **Provider-type resolution** — removed the last two hardcoded `"aws"` fallbacks. The app schema now raises `ConfigurationError` (pointing at `orb init`) when no active provider is configured rather than defaulting to a provider or reaching into the provider registry (which would breach the config-layer boundary); the explicit configured provider still wins.
- **AWS boto config DRY** — four remaining inline botocore `Config` constructions now route through the shared `get_boto3_config()` helper. Per-site values are preserved (including `aws_client`s config-driven timeouts + region); retry mode aligns with the rest of the codebase (adaptive).
- **SDK parameter mapping** — the CQRS query/command factories computed the parameter mapping twice (try body + except block); it is now computed once and reused.

### Tests
- Thread-safe rewrite of two concurrency tests that patched a shared bus mock across worker threads (a `patch.object` race that intermittently raised `AttributeError`).
- New deprecation-alias guard test: a `Template` field exposing a deprecated legacy alias must have a `model_validator(mode="before")` warning hook, so a new silent operator-facing deprecation cannot be introduced.
- Direct unit coverage for `AWSMachineAdapter._extract_status_reason` (PascalCase + snake_case shapes).
- Moto SDK lifecycle test now asserts instance state is terminating/terminated after a return cycle.
- New SQLite concurrent-write test mirroring the JSON one; storage lock scope documented as in-process, with SQLites cross-process file locking documented at the connection layer.
- Orchestrator-level tests covering the not-found flag, the no-flag-on-sync-error path, and batch partial-failure.

### Cleanup / docs / CI
- Removed the unused `RequestHistoryEvent` value object (an unwired value object with no production callers).
- MCP tool descriptions now accurately reflect handler behaviour (fixed ~14 generic first-line docstrings that surface in `tools/list`).
- Cleared ruff lint debt in the bootstrap package (dead `noqa` directives, an unnecessary quoted annotation, a redundant lambda).
- Dropped inaccurate "Deprecated" docstrings from three internal bootstrap aliases (they have live callers and emit no warning).
- Removed the manual per-PR CHANGELOG checkbox and the workflow step that nudged manual edits, now that the changelog is generated automatically from conventional commits at release time (format-validation and preview retained).

## Type of Change

## How Has This Been Tested?
- pyright: 0 errors across all changed files. import-linter: 3 contracts kept, 0 broken. ruff: no new findings on changed files (net removal of lint debt). Combined coverage gate passes at ~83%. Affected suites pass (api, sdk, aws provider, storage, mcp, integration workflow, security, e2e); the two previously-intermittent concurrency tests were run repeatedly to confirm stability.

## Security Considerations
- [x] Adds a startup warning when Host-header protection is disabled. The 404 change makes a genuinely-missing request return 404 instead of a 200 success envelope; existing requests (including failed ones) are unaffected.

## Reviewers

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -39,7 +39,6 @@ Fixes #
 - [ ] I have added tests that prove my fix is effective or that my feature works
 - [ ] New and existing unit tests pass locally with my changes
 - [ ] Any dependent changes have been merged and published
-- [ ] I have updated the CHANGELOG.md file
 - [ ] I have updated the version number (if applicable)
 
 ## Additional Notes
--- a/.github/workflows/changelog-validation.yml
+++ b/.github/workflows/changelog-validation.yml
@@ -67,34 +67,7 @@ jobs:
     - name: Validate changelog format
       run: |
         make changelog-validate
-    
-    - name: Check changelog is up to date
-      if: github.event_name == 'pull_request'
-      run: |
-        echo "Checking if changelog needs updates..."
-        
-        # Get base branch
-        BASE_SHA="${{ github.event.pull_request.base.sha }}"
-        HEAD_SHA="${{ github.sha }}"
-        
-        # Check if there are commits that should be in changelog
-        COMMIT_COUNT=$(git rev-list --count "${BASE_SHA}..${HEAD_SHA}")
-        
-        if [ "$COMMIT_COUNT" -gt 0 ]; then
-          echo "Found $COMMIT_COUNT new commits"
-          
-          # Check if changelog was modified
-          if git diff --name-only "${BASE_SHA}..${HEAD_SHA}" | grep -q "CHANGELOG.md"; then
-            echo "Changelog was updated"
-          else
-            echo "WARNING: New commits found but changelog not updated"
-            echo "Consider running: make changelog-preview"
-            # Don't fail - just warn
-          fi
-        else
-          echo "No new commits to check"
-        fi
-    
+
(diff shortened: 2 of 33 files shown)


canonicalname (approved):
LGTM

PR #344, opened 2026-07-27
chore: remove unused k8s-legacy constraints.txt

fgogolli:
## Description
Removes `k8s-legacy/constraints.txt`, a leftover pinned-dependency file from the pre-consolidation vendored k8s-legacy tree.

**Why:** nothing installs from it — no CI job, build step, Dockerfile, or Makefile references it, and its pins are already governed by the root project's dependency extras and lockfile. Because the file is a valid pip manifest of `==`-pinned packages, Dependabot's recursive manifest discovery kept opening update PRs against it despite it being dead code. Removing it stops that PR noise; it carries no dependency information used anywhere else.

Scope is deliberately limited to the single dead file. Full retirement of the remaining vendored top-level `k8s-legacy/` scaffolding (relocating the still-live HostFactory shell scripts, then removing the rest) is tracked separately as its own effort, since that touches an operator-documented deployment path.

## Type of Change

## How Has This Been Tested?
- No test needed: the file is not referenced by any build, CI, Dockerfile, or Makefile (grep-verified). The live `src/orb/k8s_legacy/` plugin and its CI leg are unaffected.

## Security Considerations

## Reviewers


canonicalname (approved):
Approved as discussed.

PR #350, opened 2026-07-28
fix(health): split liveness/readiness and stop optional providers forcing 503

fgogolli:
## Description

`GET /health` returned HTTP 503 whenever **any** enabled provider's connectivity check failed. That check reports "unhealthy" when a provider API is unreachable, which forced the aggregate `/health` status to "unhealthy" for the whole service. In a multi-provider deployment — e.g. a primary AWS provider alongside a secondary Kubernetes provider pointed at an unreachable cluster — a single unreachable optional provider took the entire endpoint down, so any consumer gating on `/health` (including a Kubernetes probe) treated the service as never-ready even though the core service was fully functional.

This change separates the concerns behind a single overloaded endpoint:

1. **Provider connectivity is now a degraded signal, not a hard failure.** An unreachable provider API — STS / EC2 / Kubernetes API, including botocore connection errors (e.g. endpoint-connection / connect-timeout) as well as API errors — reports "degraded" rather than "unhealthy". Since `/health` already maps degraded to 200, an unreachable provider no longer forces 503. Genuine core-dependency failures are unaffected.

2. **New `/livez` and `/readyz` probe endpoints.** `/health` is kept unchanged for back-compat and full detail.
   - `/livez` (liveness) returns 200 whenever the process is up and never consults any dependency.
   - `/readyz` (readiness) returns 503 when a **core** dependency is not fully healthy, otherwise 200. Storage / database is always core. Provider connectivity does **not** gate readiness in a **multi-provider** deployment — a dead secondary provider (or even the default) leaves `/readyz` at 200 so the replica stays in rotation. **But when the default provider is the sole enabled provider**, its connectivity check is classified core, so a single-provider deployment whose only provider is unreachable correctly reports not-ready (`/readyz` 503). Because a connectivity failure surfaces as "degraded", `/readyz` treats a degraded core dependency as not-ready as well as an unhealthy one.

3. **Provider connectivity checks are scoped to the default/active provider instance.** Previously a check was registered for every enabled instance, so a secondary provider could register a probe that dragged the aggregate status down. Registration is now gated on the default/active instance (explicit `default_provider_instance`, otherwise the first enabled instance), matching provider-selection precedence.

To support readiness gating, health checks are classified by kind at registration time (`core`, `provider`, `system`); readiness is derived from `core` checks only. The default provider's connectivity is classified `core` only when it is the sole enabled provider, and `provider` otherwise.

## Type of Change

## Related Issues

Fixes the case where a single unreachable provider forced `/health` to 503 for the whole service.

## How Has This Been Tested?

New unit tests cover: `/livez` returning 200 even when a dependency is unhealthy; `/readyz` returning 200 when only a provider-connectivity check fails in a multi-provider setup and 503 when a core dependency (database) is unhealthy; the sole-vs-multi readiness gating — a sole enabled provider's connectivity classified core so an unreachable sole provider makes `/readyz` not-ready, while a dead secondary (or the default) in a multi-provider setup does not; provider-connectivity failures mapping to "degraded" for the AWS (`aws`, `ec2` — including a botocore connection error, not only an API error) and Kubernetes checks; `/health` retaining its full detailed body and existing status mapping; and connectivity-check registration being scoped to the default/active instance so a secondary enabled instance registers no check. Provider clients are mocked to simulate an unreachable API — no live cloud access required.

## Test Configuration
* Python version: 3.12+
* OS: macOS / Linux
* AWS region: n/a (unit tests mock provider clients)
* Dependencies changed: none

## Checklist

## Additional Notes

The readiness classification is explicit rather than name-based: `register_check` takes a `kind` argument (`core` / `provider` / `system`). Storage/database registers as `core`. The AWS (`aws`, `ec2`) and Kubernetes (`kubernetes_api`) connectivity probes register as `core` when their provider is the sole enabled instance and `provider` otherwise, so only a single-provider deployment gates readiness on provider connectivity. Instance scoping fails open — when the instance name or provider config cannot be resolved, the check still registers so single-provider deployments keep their probe.

## Performance Impact

## Security Considerations

`/livez` and `/readyz`, like `/health`, are exempt from authentication so orchestrators can poll them without credentials — standard for Kubernetes liveness/readiness probes. They expose no sensitive detail: `/livez` returns only `{"status": "alive"}`, and `/readyz` returns only the readiness status derived from core dependency checks. Both are also allowed under read-only mode and skipped by the audit log, matching the existing treatment of `/health`.

## Deployment Notes

Operators should point their Kubernetes probes at the new dedicated endpoints instead of `/health`:

- `livenessProbe` → `GET /livez` — succeeds whenever the process is up; use it to detect a wedged process, not downstream outages.
- `readinessProbe` → `GET /readyz` — takes a replica out of rotation when a core dependency is down. Core is storage/database always, plus the sole provider's connectivity in a single-provider deployment. In a multi-provider deployment an unreachable secondary (or default) provider does not remove the replica from service.

Example manifest fragment:

```yaml
livenessProbe:
  httpGet:
    path: /livez
    port: 8000
  initialDelaySeconds: 5
  periodSeconds: 10
readinessProbe:
  httpGet:
    path: /readyz
    port: 8000
  initialDelaySeconds: 5
  periodSeconds: 10
```

`/health` remains available and unchanged (full detailed body) for dashboards and manual inspection.

## Reviewers

### DIFF
--- a/sdk/csharp/src/FINOS.OpenResourceBroker/OrbClient.cs
+++ b/sdk/csharp/src/FINOS.OpenResourceBroker/OrbClient.cs
@@ -326,8 +326,46 @@ public sealed class OrbClient : IAsyncDisposable
     }
 
     // ---------------------------------------------------------------------------
-    // System / Observability — 4 operations
+    // System / Observability — 6 operations
     // ---------------------------------------------------------------------------
+
+    /// <summary>livenessCheck — GET /livez
+    ///
+    /// The liveness probe returns 200 whenever the process is up and never gates
+    /// on any dependency or provider, so it is safe to poll as a container
+    /// liveness check.</summary>
+    public async Task<Dictionary<string, object?>> LivenessAsync(CancellationToken ct = default)
+        => await GetAsync<Dictionary<string, object?>>("/livez", ct: ct).ConfigureAwait(false);
+
+    /// <summary>readinessCheck — GET /readyz
+    ///
+    /// The readiness probe returns 200 unless a core dependency (storage/database)
+    /// is unhealthy, in which case it returns 503. A 503 carries a valid not-ready
+    /// body and is accepted as data rather than an error (mirroring HealthAsync),
+    /// so status-based retry is skipped and the not-ready status is returned
+    /// directly.</summary>
+    public async Task<Dictionary<string, object?>> ReadinessAsync(CancellationToken ct = default)
+    {
+        CheckHealth();
+        using var req = new HttpRequestMessage(HttpMethod.Get, "/readyz");
+        req.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        AddSchedulerHeader(req);
+        // 503 (not-ready) is a normal steady-state result we accept as valid data
+        // below; skip status-based retry so it returns immediately. Network-error
+        // retry is intentionally left intact.
+        req.Options.Set(RetryDelegatingHandler.SkipStatusRetryOption, true);
+        var resp = await _http.SendAsync(req, HttpCompletionOption.ResponseContentRead, ct).ConfigureAwait(false);
+        if (resp.StatusCode != System.Net.HttpStatusCode.OK &&
+            resp.StatusCode != System.Net.HttpStatusCode.ServiceUnavailable)
+        {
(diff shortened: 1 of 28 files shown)


canonicalname (approved):
LGTM

PR #351, opened 2026-07-28
fix: recover review fixes (audit log, HTTPS proxy, config, --wait, SSE) + enable npm OIDC publishing

fgogolli:
## Description
Follow-up fixes for issues found reviewing the recently-merged quality work. Each is an independent correctness fix with a regression test.

### Security / logging
- **Audit trail attribution** — the structured auth audit logger emitted its records through a JSON formatter that dropped every structured field (user_id, client_ip, reason, path, method), and the audit logger inherited the app root log level so events were silently discarded when the level was above INFO. The formatter now serialises all fields and the audit logger pins its own level, so the security audit trail is attributable and independent of the app log level.
- **HTTPS behind a TLS-terminating proxy** — with `require_https` enabled behind an off-host proxy that forwards plain HTTP, the redirect middleware saw scheme `http` and produced an infinite 307 loop. A scheme-only forwarded-proto middleware, trusting only configured proxies, resolves the scheme correctly; a spoofed header from an untrusted peer is ignored and HTTPS enforcement is preserved.

### Correctness
- **`config set` persistence** — a dict-valued `set()` replaced the value in memory but deep-merged it on disk, re-introducing keys the user had replaced. Persistence now replaces at the edited path, matching in-memory semantics.
- **`--wait` resilience** — request-status `--wait` polling aborted on the first transient provider-sync error, reporting failure for a healthy request. It now tolerates a bounded number of consecutive transient errors before giving up, matching the acquire/return poll loops; a genuine terminal failure still stops immediately.

### UI
- **Live row updates** — a status-transition SSE event patched only the row status, leaving counts and the progress bar stale; the patch now refreshes the derived count/progress fields too.
- **SSE reconnect replay** — the stream assigned a per-event sequence id but never transmitted it, so a reconnect during an outage permanently lost events (a missed terminal event left a row stuck). The server now emits the id on the wire, the client resumes from the last id via `since_seq=`, and a replay-truncated signal triggers a full refresh.

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
- Each fix carries a regression test that fails on the prior behaviour and passes after. Type checks clean on all changed files; the enforced lint set is clean; the affected suites pass.

## Security Considerations
- [x] Restores auth audit-trail attribution and fixes an HTTPS-enforcement edge case; the forwarded-proto trust is limited to configured proxies.

## Reviewers

### CI / release
- **npm trusted publishing** — the SDK publish job set up Node 20, but npm OIDC trusted publishing to registry.npmjs.org requires Node >= 22.14.0 (in addition to npm CLI >= 11.5.1). On Node 20 the OIDC exchange never ran and npm fell back to (empty) token auth, failing with `ENEEDAUTH` on the trusted path — which is why automatic npm releases never published while token-based publishing worked. The publish job now uses Node 22, and logs the active node/npm versions before publishing so an auth failure is diagnosable.

### DIFF
--- a/.github/workflows/reusable-sdk-publish.yml
+++ b/.github/workflows/reusable-sdk-publish.yml
@@ -99,10 +99,15 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
 
-    - name: Set up Node.js 20 (with npm registry auth)
+    - name: Set up Node.js (with npm registry auth)
       uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020  # v7.0.0
       with:
-        node-version: '20'
+        # npm OIDC trusted publishing requires Node >= 22.14.0 (in addition to
+        # npm CLI >= 11.5.1); on Node 20 the publish never performs the OIDC
+        # exchange and falls back to token auth, failing with ENEEDAUTH on the
+        # trusted path. Token/GitHub-Packages publishing works on any Node, but
+        # the shared setup must satisfy the strictest (trusted) path.
+        node-version: '22'
         # Point npm at the destination registry so setup-node writes the matching
         # auth line into .npmrc for NODE_AUTH_TOKEN.
         #   github            -> npm.pkg.github.com  (GITHUB_TOKEN auth)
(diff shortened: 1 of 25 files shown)


canonicalname (approved):
LGTM

